PR #55, opened 2021-07-16
Add mbio phase1 apps

asizemore:
_Goal: add the five mbio phase 1 app visualizations to the data service_

# To Dos

- [ ] Split facets to be explicitly facet1 and facet2 for apps
- [ ] Enforce rules for listVars
- [ ] Accept listVar info passed from the client
- [ ] ListVars in rel abundance box and scatter only.

Recall, each viz will show computed variables, and will be opinionated about how they are shown. For more detail, see the mbio planning google doc.

# App details
## Alpha diversity
- Includes scatterplot, boxplot
- Computed variable: alpha diversity (number, continuous)
- Scatterplot: computed variable _must_ be shown on the y-axis. Otherwise no restrictions.
- Boxplot: computed variable _must_ be shown on the y-axis. Otherwise no restrictions.

## Beta diversity
- Includes scatterplot
- Computed variables: (x,y) coordinates per sample generated from dimensionality reduction.
- Scatterplot: Both axes will show computed variables. x-axis shows the first dimension, and y-coord shows the second dimension (see [example](https://microbiomedb.org/mbio/app/workspace/strategies/323635313/425936533/analysis:202279973)). Additionally, no trend lines are allowed, and no faceting is allowed. Requires gradient colorscale.

## Relative abundance
- Includes scatterplot, boxplot
- Computed variable: a ranking of taxa based on their abundance in the samples.
- Boxplot: abundance _must_ be shown on the y-axis. Otherwise no restrictions. 
- Scatterplot: abundance _must_ be shown on the y-axis. Otherwise no restrictions.

# Notes and questions
- Info such as `String yVar = toColNameOrEmpty(spec.getYAxisVariable());` may need to come from the compute service. For example, in beta diversity the axis name includes the percent variance explained, so that the x axis title might read "Axis 1 (45%)". Could be handled here or elsewhere.
- Internally I call the relative abundance app just "abundance", because we are not actually computing relative abundance from total abundance in this app. Wojtek has already done this for us. However, the display name is Relative Abundance since the user is exploring relationships between vars and relative abundance.
- The compute service will need to return info such as default ranges. For example, some computed measures are bounded within [0, 1]. 
- The abundance app (and eventually many mbio eda vizs) will need to handle listvars (see [plot.data pr](https://github.com/VEuPathDB/plot.data/pull/39)). I am working on adding this functionality to the plugins, but any advice would be appreciated!
- How do the computed vars fit into the dependency order `.dependencyOrder("yAxisVariable", "xAxisVariable", "overlayVariable", "facetVariable")`? A user will need to choose an entity in order to configure the app, so perhaps the viz dependency order depends on that choice?


asizemore:
See comments in `AbundanceBoxplotPlugin.java` for ideas around how to handle listvars. 

Currently, plot.data expects receive listvars in the form of duplicated plot references. For example
```map <- data.frame('id' = c('y', 'x', 'z', 'group'), 'plotRef' = c('xAxisVariable', 'xAxisVariable', 'xAxisVariable', 'overlayVariable'), 'dataType' = c('NUMBER', 'NUMBER', 'NUMBER','STRING'))```

To summarize, if `getVariableEntityId(spec.getXAxisVariable());` returned a string array, then we could send that array to a function that takes an array of strings and makes it into a comma separated string. So input is ['y', x', 'z'] and output is 'y, x, z'. Then we could have 
```String xVarEntity = toCommaSeparated(getVariableEntityId(spec.getXAxisVariable()));```
return the appropriate string that can be plugged in below to the `map` var sent to plot.data.

**Review by d-callan — COMMENTED** [review]
1. Id avoid the word 'OTU' not sure it's technically accurate since we use dada2
2. It's probably ok to assume that a boxplot response always looks consistent. As in, no matter how you configure a boxplot (or what the viz config is, here) they always need the same data supplied to the web component. So the responses you copied are things we can probably refactor/ reuse. Thoughts?
3. Probably compute var labels should be able to be optionally passed along from the compute service to plot.data. think that would mean a new, optional, col in our var metadata map. And a new issue in plot.data to return labels as part of var details when available. And also an update to the var details api in the data service.
4. Think these apps should showMissingness in the same way as the pass app? So you're missing some viz config..
5. I like the helper, but you might be able to name it better?

**Comment by d-callan** [comment]
@ryanrdoherty may know better, but I think the compute config needs moved to eda common?

**Comment by asizemore** [comment]
Great thank you! Updates in the recent commits.
For 1, you're right OTU isn't correct and since we might have so many abundance-type data (pathways maybe?) flowing through the app I made the descriptions vague.
2. Sounds good to me. I did the same for scatter as well.
3. I didn't finish addressing this because it'll need more global work, but offered an option in `AbundanceBoxplotPlugin.java`.
4. Yup! I had missed lots of new config so all app viz raml should be up to date with pass viz raml (and java).
5. Could definitely name it better :) How about `toCommaSepString`? Open to suggestions!

**Comment by d-callan** [comment]
two further thoughts i had.. ill use abundance box as ex. if what the user chooses is a ranking method, then they dont really need to choose an xaxis variable. the xaxis ticks are determined by the compute service. in that case, the request doesnt need the xAxisVariable. if the user chooses up to ten taxa on their own, then we would need xAxisVariable in the request and itd have to be an array. fortunately, weve talked to Dan and he seems mostly ok we moving listVars to the pass app. 
all that to say, i think we can:
1. remove xAxisVariable from abundance boxplot request (and look for other places to apply the same logic) and 
2. add pass app vizs for box and scatter w the appropriate vars able to take arrays up to 10 values

File: src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPluginWithCompute.java [hunk 1]
```diff
@@ -41,4 +59,21 @@ private void validateComputeName(String passedComputeName, String appName) {
...
+    EntityDef collectionVarEntityDef = _referenceMetadata.getEntity(collectionVar.getEntityId()).get();
+    VariableDef collectionVarDef = _referenceMetadata.getVariable(collectionVar).get();
+    TreeNode<VariableDef> childVarsTree = collectionVarEntityDef.getNativeVariableTreeNode(collectionVarDef);
+    // for now assuming we only have leaves as children. revisit if that turns out to not be true
+    Collection<VariableDef> childVarDefs = childVarsTree.flatten();
+
+    return new ArrayList(childVarDefs);
+  }
+
+  protected VariableSpec getComputeEntityPrimaryKeyVarSpec(String entityId) {
```
d-callan, line 74: think were calling these id vars, rather than primary keys?

src/main/java/org/veupathdb/service/eda/ds/plugin/alphadiv/AlphaDivBoxplotPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+          getVariableSpecFromList(spec.getFacetVariable(), 1)));
+      // merge alpha div and other viz stream data as input to boxplot
+      connection.voidEval("vizData <- merge(alphaDivDT, " + 
+                                            DEFAULT_SINGLE_STREAM_NAME + 
+                                         ", by=" + computeEntityPKColName +")");
+      connection.voidEval(getVoidEvalVarMetadataMap(DEFAULT_SINGLE_STREAM_NAME, varMap));
+      // update the new map obj in R to add alphaDiv
+      connection.voidEval("map <- rbind(map, list('id'='alphaDiversity'," +
+                                                 "'plotRef'='yAxisVariable'," +
+                                                 "'dataType'='NUMBER'," +
```
asizemore, line 136: data type and shape should probably come from computedVariableDetails. I'll add it!

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
```diff
@@ -102,7 +102,7 @@
     validateVisualizationSpec(_pluginSpec);
 
     // get list of data streams required by this subclass
-    _requiredStreams = getRequestedStreams(_pluginSpec);
+    _requiredStreams = _requiredStreams == null ? _requiredStreams = getRequestedStreams(_pluginSpec) : _requiredStreams;
```
ryanrdoherty, line 105: Not sure what you're doing here (and with the protected above).  There should still only be one call to get the required streams, yes?  Just overriding to force compute viz classes to implement the two-arg version?

d-callan: yes. thats what im trying to do anyway. i kind of figured youd pick that out.. and was prepared to say something like 'ryan its just until we have a compute service'

d-callan: if youve got a better short term solution than that, feel free, obviously

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
```diff
@@ -239,8 +239,12 @@ protected String getVariableDataShape(List<VariableSpec> vars, int index) {
   protected String singleQuote(String unquotedString) {
     return "'" + unquotedString + "'";
   }
+
+  protected String getVoidEvalFreadCommand(String fileName, VariableSpec... vars) {  
+    return getVoidEvalFreadCommand(fileName, new ListBuilder().addAll(vars).toList());
+  }  
   
-  protected String getVoidEvalFreadCommand(String fileName, VariableSpec... vars) {    
+  protected String getVoidEvalFreadCommand(String fileName, List<VariableSpec> vars) {
```
ryanrdoherty, line 247: Probably don't need both here.  If you keep the varargs one and have a list in the calling code, use List.toArray()

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPluginWithCompute.java
```diff
@@ -14,6 +24,12 @@
 
   // returns the compute-spec class this visualization expects
   protected abstract Class<R> getComputeSpecClass();
+  protected List<StreamSpec> getRequestedStreams(S pluginSpec) {
+    return getRequestedStreams(pluginSpec, _computeSpec);
+  }
+  protected abstract List<StreamSpec> getRequestedStreams(S pluginSpec, R computeConfig);
+
+  protected static final String COMPUTE_STREAM_NAME = "compute_input_dataset";
```
ryanrdoherty, line 32: My impression was that we would somehow tell the merging service how to fetch the compute result stream for possible merging with regular and derived vars rather than ask compute service for it directly.  Is that a wrong understanding?

src/main/java/org/veupathdb/service/eda/ds/plugin/alphadiv/AlphaDivBoxplotPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+          spec.getOverlayVariable(),
+          getVariableSpecFromList(spec.getFacetVariable(), 0),
+          getVariableSpecFromList(spec.getFacetVariable(), 1)));
+      // merge alpha div and other viz stream data as input to boxplot
+      connection.voidEval("vizData <- merge(alphaDivDT, " + 
+                                            DEFAULT_SINGLE_STREAM_NAME + 
+                                         ", by=" + computeEntityPKColName +")");
+      connection.voidEval(getVoidEvalVarMetadataMap(DEFAULT_SINGLE_STREAM_NAME, varMap));
+      // update the new map obj in R to add alphaDiv
+      connection.voidEval("map <- rbind(map, list('id'='alphaDiversity'," +
```
asizemore, line 134: 'id' can also come from computedVariableDetails$id. Not necessary, just a suggestion

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPluginWithCompute.java
ryanrdoherty: Oh, I see below- we are still getting them from merging in a second stream.  Ok if the subset is still applied here?

ryanrdoherty (approved):

src/main/java/org/veupathdb/service/eda/ds/plugin/alphadiv/AlphaDivBoxplotPlugin.java
asizemore, line 134: if col names should always be entity.id, then 'id' here should be toColNameOrNull(computedVariableDetails) i think. I'll make sure computedVariableDetails will work with that function

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
ryanrdoherty: Don't understand why the original line here was not sufficient- looks like you are caching the value in case you need to ask for streams multiple times.  But we aren't do that, are we?

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPluginWithCompute.java
d-callan: Yea. I think it's fine the filtering is applied. 

Right now we make two requests, one is input to the compute, the other just raw inputs for the viz. And the viz plugin actually merges the output of the compute to the input for the viz. 

In the ideal, the viz plugin makes a single request to merging, and the compute output and other viz inputs are merged there like you described. And the compute service would make its own request to merging for its own inputs, w the same filtering criteria applied.

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
d-callan: I think you're probably right. How to handle the requested streams was something I changed my mind about a couple times. But the way I have it now I think we don't need this change.

src/main/java/org/veupathdb/service/eda/ds/plugin/alphadiv/AlphaDivBoxplotPlugin.java
asizemore: Updated in microbiomeComputations. `computedVariableDetails` now contains `variableId`, `entityId`, `dataType`, `dataShape` (among others).

src/main/java/org/veupathdb/service/eda/ds/plugin/alphadiv/AlphaDivBoxplotPlugin.java
d-callan: i forgot id in plot.data means the name of a column. we should probably clarify that.
this is also interesting bc i think it means moving toColNameOrNull to veupathUtils, and making sure veupathUtils is installed in our rserve container.

src/main/java/org/veupathdb/service/eda/ds/plugin/alphadiv/AlphaDivScatterplotPlugin.java
```diff
@@ -134,7 +134,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
           DEFAULT_SINGLE_STREAM_NAME + 
        ", by=" + computeEntityIdColName +")");
       connection.voidEval(getVoidEvalVarMetadataMap(DEFAULT_SINGLE_STREAM_NAME, varMap));
-      connection.voidEval("map <- rbind(map, list('id'=attributes(alphaDivDT)$computedVariableDetails$variableId," +
+      connection.voidEval("map <- rbind(map, list('id'=veupathUtils::toColNameOrNull(attributes(alphaDivDT)$computedVariableDetails)," +
                                                  "'plotRef'='yAxisVariable'," +
                                                  "'dataType'=attributes(alphaDivDT)$computedVariableDetails$dataType," +
                                                  "'dataShape'=attributes(alphaDivDT)$computedVariableDetails$dataShape," +
```
asizemore, line 140: Note alphaDiv and betaDiv are also returning displayLabels in that computedVariableDetails list.

asizemore: `attributes(dt)$computedVariableDetails$displayLabel`

asizemore (commented):
For now we can probably just hard code axis ranges to 0,1, but when we need them the default ranges are in `attributes(alphaDivDT)$computedVariableDetails$defaultRange`

d-callan:
im a little concerned were conflating two ideas wrt computed vars: 1) variable metadata and 2) variable details
the first are things like display labels and ranges used by the front end i think and the second is essentially a collection of things plot.data needs to know to identify a variable. 
im starting to feel inclined to making them available separately of each other. that way one can get routed to rserve and the other basically passed along to the client without having to take a circuitous route.

asizemore:
ahhh yes we are. 
Hmm does plot.data need displayLabels as inputs at all? Originally we added them for sake of these apps, but if the apps can return displayLabels and send them directly to the client, maybe plot.data doesn't need them at all?

I'll break up the current details into variableMetadata and variableDetails here and in microbiomeComputations

d-callan:
was just talking w dave, he thinks we can optionally include this stuff in the config section of viz responses. i think that makes it easier to have the vizs be reusable in web-eda. so while were using plot.data to put those together, we'll have to pass it this info i suppose, and make sure the plot config matches the api defined here.

d-callan:
we also talked about adding something like 'options' to the app metadata where we can declare things like 'this viz doesnt want trend lines even though usual scatter plots do'

d-callan:
@ryanrdoherty @asizemore i *think* the only thing remaining is to add something to the apps metadata so that we can specify for ex when a particular scatter plot viz wants trends lines or not. anything ive forgotten?

d-callan:
probably clarify collection var api too...

ryanrdoherty:
Looks pretty good to me.  You planning to merge to master?

d-callan:
not quite yet, but very soon

PR #58, opened 2021-07-21
add paging to pass app table plugin

d-callan:


src/main/java/org/veupathdb/service/eda/ds/plugin/pass/TablePlugin.java
```diff
@@ -101,10 +105,30 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
...
-      out.write(new JSONArray(row).toString().getBytes());
+      if (offset != null) {
+        if (offsetCount < offset) {
+          offsetCount++;
+          s.nextLine();
+        } else {
+          if (numRows != null) {
+            if (rowCount < numRows) {
+              rowCount++;
+              String[] row = parser.parseLineToArray(s.nextLine());
```
ryanrdoherty, line 119: Maybe break these three lines into a function?  Repetition.

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/TablePlugin.java
```diff
@@ -101,10 +105,30 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
...
+              if (first) first = false; else out.write(",".getBytes());
+              out.write(new JSONArray(row).toString().getBytes());
```
ryanrdoherty, line 121: To avoid using getBytes() everywhere and (more importantly) improve performance, wrap the passed OutputStream in a BufferedWriter and write to that instead.  e.g.
BufferedWriter bufOut = new BufferedWriter(new OutputStreamWriter(out));
bufOut.write(someString);

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/TablePlugin.java
d-callan: i agree its a bit repetitive.. and maybe im just being thick, but i dont really get what you mean by baking conditional logic into a function?

Issue #49, opened 2021-06-29
update data service to use new Variable properties


steve-fischer-200:
for convenience, here is pertinent RAML

```
   API_StringVariable:
    type: API_Variable
    discriminatorValue: string
    properties:
      distinctValuesCount: number
      isTemporal: boolean
      isFeatured: boolean
      isMergeKey: boolean
      isMultiValued: boolean


  API_NumberVariable:
    type: API_Variable
    discriminatorValue: number
    properties:
      units: string
      precision: number
      displayRangeMin?: number
      displayRangeMax?: number
      rangeMin?: number
      rangeMax?: number
      binWidthOverride?: number
      binWidth?: number
      distinctValuesCount: number
      isTemporal: boolean
      isFeatured: boolean
      isMergeKey: boolean
      isMultiValued: boolean

  API_DateVariable:
    type: API_Variable
    discriminatorValue: date
    properties:
      displayRangeMin?: string
      displayRangeMax?: string
      rangeMin?: string
      rangeMax?: string
      binWidthOverride?: string
      binWidth?: string
      distinctValuesCount: number
      isTemporal: boolean
      isFeatured: boolean
      isMergeKey: boolean
      isMultiValued: boolean

  API_LongitudeVariable:
    type: API_Variable
    discriminatorValue: longitude
    properties:
      distinctValuesCount: number
      isTemporal: boolean
      isFeatured: boolean
      isMergeKey: boolean
      isMultiValued: boolean

```

ryanrdoherty:
Danielle has only requested range and displayRange min and max, and she can now get them for date and number vars.  I will add isMultiValued in anticipation of needing it to parse those out of tabular streams.

PR #65, opened 2021-07-30
Backend: performance improvements after profiling w umsp

d-callan:
So far the primary change is to declare column types to R's data.table::fread fxn so that it doesnt have to waste time figuring them out


asizemore (approved):
I can't comment on the java specifics, but I like the organization and it's really easy for me to follow.

SheenaTomko:
Touched base with Danielle and main things to qa here and in #49 are 1) do things seem to still be working despite some fairly significant refactoring of the code and 2) does it take less time to see plots. the main offender was the scatter plot in umsp. there were a few requests there (mostly involving various combinations of age, date, weight and height) that were taking fairly significantly >1min. if we have most things under a minute, and age vs date around a minute, id say we were successful

UMSP is currently being reloaded, so will address on Tuesday

SheenaTomko:
Didn't notice any issues with other studies when qc'ing other visualization-related tickets. 

Tested UMSP. Non-scatter plots (bar charts, histograms, etc.) took ~40 sec to load with and without overlay variables while a scatter plot of date vs age took 2 min. Date vs height with or without sex overlaid took 1.5 min. Not quite as fast as hoped for, but there's a separate [ticket ](https://github.com/VEuPathDB/web-eda/issues/291) to add a warning when the wait time will be long since there's lots of study data. This should be sufficient, so marking this as complete.

PR #81, opened 2021-09-06
Update viz display names

asizemore:
With web-eda [PR #379](https://github.com/VEuPathDB/web-eda/pull/379) , addresses web-eda [issue #325](https://github.com/VEuPathDB/web-eda/issues/325#issuecomment-912742672)

This PR makes the following updates to visualization display names:
- "RxC Contingency Table" to "Mosaic plot, RxC Table"
- "2x2 Contingency Table" to "Mosaic plot, 2x2 Table"
- "Line Plot" to "Time series"

**NOTE**: The above update to the line plot plugin assumes that eventually we'll need to make a new plugin for mbio pass app - recall mbio would like a regular line plot for their pass app and clinepi only wants a time series. I am guessing that when the time comes we'll split the line plot plugin into two because there will be different var constraints?


d-callan (approved):

d-callan:
might be good to update the line plot description too

PR #89, opened 2021-09-12
Mods to support user service integration

ryanrdoherty:
Tried to write the service info into docker-compose based on other services and prior discussion (i.e. routing /users/ requests to the new service), but I wouldn't want to call this "complete" and cannot test locally on my laptop because of LDAP/ssh-tunneling limitations.


rbelnap (approved):
just needed a bit of a tweak to the traefik rule

PR #101, opened 2021-09-24
limit-request-sizes

d-callan:
related to veupathdb/web-eda#324


d-callan:
the chosen limits are slightly arbitrary right now. i think id call them slightly educated guesses. wont be able to fiddle w boxplot when all points on, bc i think we dont support that feature on the frontend yet? and umsp doesnt appear to be loaded currently to test scatter? anyhow, i can adjust later if necessary.

d-callan:
limits now supplied by client, and were counting the number of plottable rows (based on the value of showMissingness). not yet tested in any way.

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/BoxplotPlugin.java
```diff
@@ -101,18 +103,31 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
...
+    nonStrataVarColNames.add(toColNameOrEmpty(spec.getXAxisVariable()));
+    nonStrataVarColNames.add(toColNameOrEmpty(spec.getYAxisVariable()));
+
+    RFileSetProcessor filesProcessor = new RFileSetProcessor(dataStreams)
+      .add(DEFAULT_SINGLE_STREAM_NAME, 
+        spec.getMaxAllowedDataPoints(), 
+        showMissingness, 
+        nonStrataVarColNames, 
+        (name, conn) ->
+        conn.voidEval(getVoidEvalFreadCommand(DEFAULT_SINGLE_STREAM_NAME,
```
ryanrdoherty, line 118: A little cleaner here to use the parameter 'name' than the constant again; works, but makes it less clear that we could potentially use this bifunction elsewhere (with other file names).

src/main/java/org/veupathdb/service/eda/ds/util/RFileSetProcessor.java
```diff
@@ -18,40 +19,56 @@
...
   }
 
   public RFileSetProcessor add(String name, BiConsumerWithException<String, RConnection> fileReader) {
-    return add(name, null, fileReader);
+    return add(name, null, "FALSE", null, fileReader);
   }
 
-  public RFileSetProcessor add(String name, Integer maxAllowedRows, BiConsumerWithException<String, RConnection> fileReader) {
+  public RFileSetProcessor add(String name, Integer maxAllowedRows, String showMissingness, List<String> nonStrataColNames, BiConsumerWithException<String, RConnection> fileReader) {
+    Boolean booleanShowMissingness = showMissingness.equals("FALSE") ? false : true;
```
ryanrdoherty, line 50: A little more NPE-safe to use:
"FALSE".equals(showMissingness)
which will return false if showMissingness is null rather than throw NPE

src/main/java/org/veupathdb/service/eda/ds/util/RFileSetProcessor.java
```diff
@@ -18,40 +19,56 @@
...
       this.name = name;
       this.stream = stream;
       this.maxAllowedRows = processingInfo.map(i -> i.getFirst()).orElse(Optional.empty());
+      this.showMissingness = showMissingness;
+      this.nonStrataColNames = nonStrataColNames;
       this.fileReader = conn -> processingInfo.map(i -> i.getSecond()).orElse((a,b) -> {}).accept(name, conn);
     }
   }
 
   private final Map<String,InputStream> _dataStreams;
```
ryanrdoherty, line 36: I would put all this stuff in an object and use a single map; before when there were only two things I used a TwoTuple, but with four we should just make a little container object and put it in a single map (so there would be two total maps: one for InputStreams, one for config)

ryanrdoherty (approved):
Works; see comments.  If you don't make any changes, I may go in later and edit based on my comments, but this is fine to merge.

src/main/java/org/veupathdb/service/eda/ds/util/RFileSetProcessor.java
d-callan: im not opposed. i even considered it when i was working on this fri. i guess i didnt understand why it was a TwoTuple to begin. (im sure theres a reason, i just dont know it.) why couldnt/shouldnt these each just be their own map, with four (showMissingness, nonStrataColNames, maxAllowedRows, fileReader) total currently?

d-callan: also, my son came home sick from school today and my time is limited. given we already have something implemented, do you think its worth merging without this? or do you know if the front end guys already have enough to work with if i sit on this for a bit longer?

Issue #124, opened 2021-10-20
Hitting 500 errors and other errors on ELICIT


SheenaTomko:
If I try to look at ELICIT on qa.restricted.clinepidb.org, I can click through variables at first. However, after a couple of minutes, I hit the error message below. Once I hit it for one variable, I hit it for all variables, even ones that looked fine just 2 sec ago. 
<img width="778" alt="Screen Shot 2021-10-20 at 8 17 36 AM" src="https://user-images.githubusercontent.com/54597319/138100838-09f9b4a7-2a3d-46d1-8bf9-193f50431669.png">

If I try and clear my history with command+shift+R, I get the following error message for the study:

<img width="714" alt="Screen Shot 2021-10-20 at 8 18 29 AM" src="https://user-images.githubusercontent.com/54597319/138101007-e665348e-47ea-46eb-a856-69fe5bc3b5c5.png">

d-callan:
i cant reproduce. possible this has already been resolved through another ticket in the two weeks since it was reported? @SheenaTomko can you check and close if you also cant reproduce?

SheenaTomko:
This is now fixed. Closing

PR #138, opened 2022-01-05
Allow numbers, strings, and integers for overlays

asizemore:
With plot.data [PR #124](https://github.com/VEuPathDB/plot.data/pull/124), addresses backend needs for allowing a gradient colormap.

Specifically, this PR allows all numbers, integers, and strings to be overlay variables
**Important** my logic isn't perfect yet! We can't just allow _all_ strings and numbers. We have to allow all numbers, then also strings with < 9 values. Not sure how to do this yet. Keeping as draft until we figure it out!


d-callan (approved):
seems fine, though it makes me wonder if the descriptions should belong outside the individual patterns somehow rather than be repeated across them. something we can worry about later if it becomes a problem i suppose

chowington:
@d-callan Let me know when the changes are deployed on the backend---I don't know what the process is for this repo

dmfalke:
@chowington this should be available on a dev site (dfalke-b). Let me know if you run into issues with that.

d-callan:
this isnt intended for b60 right? so we may have to revert that merge...

PR #143, opened 2022-01-25
Initial apps run

asizemore:
Fixes little issues we had in the plugins.

Verified working with clinepi data. When mbio is back online, i'll test with that as well.


src/main/java/org/veupathdb/service/eda/ds/plugin/abundance/AbundanceBoxplotPlugin.java
```diff
@@ -110,30 +110,32 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
       connection.voidEval(getVoidEvalFreadCommand(COMPUTE_STREAM_NAME,
         computeInputVars
       ));
-      connection.voidEval("abundanceDT <- rankedAbundance(" + COMPUTE_STREAM_NAME + ", " + 
-                                                      computeEntityIdColName + ", " + 
-                                                      method + ")");
+
+      connection.voidEval("abundanceDT <- rankedAbundance(" + COMPUTE_STREAM_NAME + ", '" + 
+                                                      computeEntityIdColName + "', '" +
```
d-callan, line 115: i think alternatively we have a singleQuote helper somewhere

d-callan (approved):

asizemore:
After using singleQuotes, confirmed all four vizs working

PR #144, opened 2022-01-26
moved all compute config to compute config

asizemore:
Collect all compute configuration into one `computeConfig` instead of having it spread out in the plot spec as well. 

Changes affect the boxplot and scatterplot for the alpha div and ranked abundance apps.


asizemore:
Changes compile without failures and the four endpoints (box and scatter for abundance and alpha div) confirmed working with clinepi data.

d-callan (approved):

Issue #70, opened 2021-08-09
Backend: need a pie plot endpoint (for map visualizations)


d-callan:
this will call plot.data::bar, except it will have different variable constraints and no overlay var, barmode etc. dont know about proportions option. think its needed for the map.

d-callan:
some comments from bob: 
This will be very similar to a barplot endpoint, but when used in the map visualization the variable constraints will need to allow an effectively unlimited number of facets. At the same time, overlay strata should be disallowed.

The constraints on the main variable are an interesting one too... In MapVEu 1.0 we had a limited number of colours (20) and then used varying shades of grey for the remaining pie segments. So t[here](https://veupathdb.github.io/web-components/?path=/story/map-donut-markers--all-in-one-request) was no constraint on the number of values that categorical values took. This is the one benefit of the pie/donut over bar plots - the "other" categories can be lumped together effectively as one grey pie. The downside to allowing unlimited pie segments is that the legend can become very unwieldy... I would be happy if the back end aggregated all the "others" into one pie slice (as is demoed with static data here). So yeah - we do need a new endpoint for this, @d-callan, as you suggested all along!!

Maybe call it geo-pie or something, in case we want a regular pie endpoint at a later date?

d-callan:
@bobular in general in the eda cases like you describe for the main variable having more than 8 categories we would simply constrain against that term and direct the user to make a derived variable when we have them.. is there a reason thats not viable here?

SheenaTomko:
Have pie plots on maps

PR #150, opened 2022-02-11
Pieplot

d-callan:
this is needed for the map initially, but ive written it fairly generally so far in case we see other use cases in the future.

resolves veupathdb/edadataservice#70
depends on veupathdb/plot.data#140
depends on veupathdb/plot.data#141

there is still an open question re restricting the main/ xaxis var to 8 values (since thats what we have colors for) vs allowing any number of values and aggregating those not belonging to the 'top 8' ..


d-callan:
possible @bobular may want a look too, to be sure it meets his needs.

schema/url/pass/plugin-pieplot.raml
```diff
@@ -0,0 +1,49 @@
...
+  PieplotSpec:
+    additionalProperties: false
+    properties:
+      outputEntityId:
+        type: string
+      showMissingness:
+        type: string
+        enum: ['TRUE', 'FALSE']
+        required: false
+      xAxisVariable:
```
asizemore, line 20: i've been back and forth here. xAxis makes sense since that's where this var is mapped to in the bar plot. Overlay also makes sense because we're using color to stratify 'counts of participants' by this var. Any other reasons i'm missing that made xAxis more appropriate than overlay here?

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/PieplotPlugin.java
```diff
@@ -0,0 +1,102 @@
...
+      
+    useRConnectionWithRemoteFiles(dataStreams, connection -> {
+      connection.voidEval(getVoidEvalFreadCommand(DEFAULT_SINGLE_STREAM_NAME, 
+          spec.getXAxisVariable(),
+          getVariableSpecFromList(spec.getFacetVariable(), 0),
+          getVariableSpecFromList(spec.getFacetVariable(), 1)));
+      connection.voidEval(getVoidEvalVarMetadataMap(DEFAULT_SINGLE_STREAM_NAME, varMap));
+      String cmd =
+          "plot.data::bar(" + DEFAULT_SINGLE_STREAM_NAME + ", map, " +
+              "'count', 'group', " +
```
asizemore, line 97: Are we planning to eventually allow the user to switch between 'count' and 'proportion'?

asizemore (approved):
looks great, just had some general questions

schema/url/pass/plugin-pieplot.raml
```diff
@@ -0,0 +1,49 @@
...
+      config: PieplotSpec
+  
+  PieplotSpec:
+    additionalProperties: false
+    properties:
+      outputEntityId:
+        type: string
+      showMissingness:
+        type: string
+        enum: ['TRUE', 'FALSE']
```
asizemore, line 18: is there a plan for how to represent missingness? Not in the tiny pie view but maybe in the expanded pie view? Like hashing or something? We don't have markers to help distinguish the gray from other colors here.

schema/url/pass/plugin-pieplot.raml
d-callan: Idk, I guess in my head it's just a polar x-axis.. but we could just call it 'main' variable or something too of that's clearer generally

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/PieplotPlugin.java
d-callan: I think I initially assumed we might and then decided that it probably wasn't very useful.. Its a plot meant to visually support part -to-whole relationships based on counts so it seemed like a pie in proportion mode would probably look the same but have less info cause it was no longer aware of the underlying counts. Am I crazy?

schema/url/pass/plugin-pieplot.raml
d-callan: None I know of.. makes the x-axis v overlay question interesting though particularly wrt consistency w other stats charts.. we usually don't provide no data counts for the axes...

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/PieplotPlugin.java
asizemore: No you're totally right it would look exactly the same just have less info :) In general i agree with you the proportion is not really useful. However, in the specific case of our specific audience, it seems they often like to see both? Thinking about the bar plot also - it has both proportion and count which, when no strata vars are selected, just changes the vertical axis.

Anyways, i'd be happy to only release "count" and wait for someone to ask for "proportion" explicitly

d-callan: I think I can just add it in case. I don't think it costs us any real work.

d-callan:
think weve decided in the dataviz meeting that if a variable has >8 values then the top 7 will get assigned their own colors and the 8th color will go to an 'other' category. gray will remain dedicated to 'no data'. at some point in the future wed like to allow users to easily and intuitively recode variables which have >8 unique values so that they can make their own 8 categories.

one question i realize we left unanswered.. is it always 'top 7' by count? its not top 7 by mean or median or something right? though i suppose that could be a type of user-defined recoding in the future too..

bobular:
Good question @d-callan - I'm pretty sure we want top N by count if we're returning counts.

Would we be returning anything else in the near future?  Do you have plans for median/mean as in lineplot? I think that's potentially dodgy from a dataviz perspective because pie/donut plots don't lend themselves to showing error bars.  I'd be happy to stick with count-based pie/donuts for categorical vars only.

d-callan:
no the only other thing i to potentially offer are proportions, as ann pointed out earlier. think thats still six of one half dozen of the other in terms of ranking the top N.

bobular also pointed out that the response currently says 'barplot' rather than 'pieplot'

only other thing ill point out is that well be finding the top N across the whole set before faceting, then actually doing the grouping by overlay per facet. that gives us the consistency we need in the color palette, but also means that its very possible the top N for a particular facet panel/ map marker will in practice be a very small total slice of the pie.

d-callan:
@asizemore maybe give this another look, and take note that the recent decisions have made it necessary to do some work in plot.data. ill have a PR for that out asap.

d-callan:
wrt providing the list of ranked values.. worth keeping in mind i can only report and rank based on what i've seen. id propose i just return the list of values ive seen if 8 or fewer, and the list of the top 7 plus explicitly list the 'other' category otherwise. there are some questions down that path though..
if a variable has >8 values by the metadata but exactly 8 or fewer after filtering then i can only provide the list of values ive seen unranked. the legend could maybe show the values seen w their own unique colors only? its just inconsistent w other vizs.. but i guess pie is already inconsistent in other ways..

schema/url/pass/plugin-pieplot.raml
```diff
@@ -0,0 +1,54 @@
...
+        required: false
+      overlayVariable: 
+        type: VariableSpec
+      facetVariable: 
+        type: VariableSpec[]
+        maxItems: 2
+        required: false
+      valueSpec: 
+        type: ValueSpec
```
asizemore, line 28: > it does seem to make intuitive sense to have the two (valueSpec and ranking method) be consistent if its not going to be explicitly user defined

I didn't quite follow - does this mean we're planning to always use the same ranking method? So no user-chosen ranking method? Totally fine with me, just making sure we don't need a `rankingMethod` property here.

schema/url/plots.raml
```diff
@@ -24,6 +24,24 @@ types:
...
+        type: StrataVariableDetails
+      facetVariableDetails:
+        type: StrataVariableDetails[]
+        maxItems: 2
+        required: false
+      value:
+        type: number[]
+        minItems: 1
+        required: true
+      label:
```
asizemore, line 40: Not sure if `label` is necessary? Would this be duplicated in the overlayVariableDetails?

asizemore, line 40: Not sure if `label` is necessary? Would this be duplicated in the overlayVariableDetails?

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/PieplotPlugin.java
```diff
@@ -0,0 +1,97 @@
...
+    return PieplotSpec.class;
+  }
+
+  @Override
+  public ConstraintSpec getConstraintSpec() {
+    return new ConstraintSpec()
+      .dependencyOrder("overlayVariable", "facetVariable")
+      .pattern()
+        .element("overlayVariable")
+          .description("Variables with more than 8 values will assign the top 7 values by count to their own categories and assign the additonal values into an 'other' category.")
```
asizemore, line 49: Just a note - we should also describe this elsewhere, maybe in the map docs. If i remember correctly i think this description only shows if the variable is disabled, which would not be the case for us

asizemore, line 49: Just a note - we should also describe this elsewhere, maybe in the map docs. If i remember correctly i think this description only shows if the variable is disabled, which would not be the case for us

asizemore (commented):
Just a few questions from me!

schema/url/plots.raml
d-callan: youre right. and value wouldnt need to be an array either. @bobular are you sure this is what you want?

schema/url/pass/plugin-pieplot.raml
d-callan: think thats exactly what were saying. or at the very least if we do choose to offer a user-defined ranking method at some point we can figure it out then

d-callan:
i think id like to propose the following as a path forward:

leave the x axis variable an x axis variable. i maintain that conceptually its a polar axis, and think this will give us the response formatted as wed want. also, change the values for showMissingness from T/F to something like 'all vars', 'strata vars' and 'no vars'. @asizemore @bobular thoughts?

d-callan:
I think we would change it for all plots. Apparently the current proposal is to do histograms in maps as well w a 'no data' bin ...

schema/url/plots.raml
bobular: Is this issue moot if we stay with using x-axis-variable as mentioned below?

d-callan: yes

d-callan:
related to veupathdb/plot.data#138 and veupathdb/plot.data#139

d-callan:
@asizemore @bobular if this looks now finally like what we think we want ill start on the plot.data work.

asizemore:
Looks good to me!

d-callan:
milestone! were returning rankedValues. :smile:  endpoint works so long as you dont pass it a showMissingness prop currently.

schema/url/pass/plugin-pieplot.raml
meshl5555: Shah

d-callan:
one (hopefully!) final look please @asizemore , after much ado

schema/url/pass/plugin-barplot.raml
```diff
@@ -14,8 +14,7 @@ types:
       outputEntityId:
         type: string
       showMissingness:
-        type: string
-        enum: ['TRUE', 'FALSE']
+        type: ShowMissingness
```
asizemore, line 17: This only needs to be showMissingnessNoAxes, no? If the plan is to use the histo plot.data function then only pie and histo would use ShowMissingness as opposed to ShowMissingnessNoAxes I'd think

schema/url/pass/plugin-heatmap.raml
```diff
@@ -14,8 +14,7 @@ types:
       outputEntityId:
         type: string
       showMissingness:
-        type: string
-        enum: ['TRUE', 'FALSE']
+        type: ShowMissingness
```
asizemore, line 17: Why not ShowMissingnessNoAxes here?

schema/url/pass/plugin-histogram.raml
```diff
@@ -13,8 +13,7 @@ types:
       outputEntityId:
         type: string
       showMissingness:
-        type: string
-        enum: ['TRUE', 'FALSE']
+        type: ShowMissingnessNoAxes
```
asizemore, line 16: this one i expected to see ShowMissingness. Sorry i'm not following 😄 What's the criteria for ShowMissingness vs ShowMissingnessNoAxes?

schema/url/plots.raml
```diff
@@ -161,14 +161,15 @@ types:
     properties:
       config: MosaicSpec
 
+  # TODO consider showMissingness here. for 2x2 if we allow axes vars it wont be 2x2 any longer...
+  # in practical terms that means the associated stats dont work. for now not allowing in either mosaic
```
asizemore, line 165: agreed with keeping it ShowMissingnessNoAxes. I don't think the stats would make sense in the RxC case. I suppose we could turn off stats like we do in some other no data cases but...

schema/url/visualizations.raml
```diff
@@ -57,15 +57,23 @@ types:
...
   # Assuming computed variable must be on y axis and that x axis may be determined by collection var or specified.
   BoxplotWith1ComputeSpec:
     additionalProperties: false
     properties:
       outputEntityId:
         type: string
       showMissingness:
-        type: string
-        enum: ['TRUE', 'FALSE']
+        type: ShowMissingness
```
asizemore, line 76: regular boxplot spec has ShowMissingnessNoAxes so that should probably also be the case here, too?

schema/url/visualizations.raml
```diff
@@ -94,8 +102,7 @@ types:
...
```
asizemore, line 105: same comment as for boxplotWith1ComputeSpec

src/main/java/org/veupathdb/service/eda/ds/plugin/abundance/AbundanceBoxplotPlugin.java
```diff
@@ -133,8 +134,8 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
       String command = "plot.data::box(vizData, map, '" +
           spec.getPoints().getValue() + "', " +
           showMean + ", " + 
-          computeStats + ", " + 
-          showMissingness + ", " +
+          computeStats + ", '" + 
+          deprecatedShowMissingness + "'', " +
```
asizemore, line 138: only 1 quote needed on this line i think

asizemore: also noticed this in the other plugins

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/PieplotPlugin.java
```diff
@@ -93,9 +90,10 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
           getVariableSpecFromList(spec.getFacetVariable(), 1)));
       connection.voidEval(getVoidEvalVarMetadataMap(DEFAULT_SINGLE_STREAM_NAME, varMap));
       String cmd =
-          "plot.data::bar(" + DEFAULT_SINGLE_STREAM_NAME + ", map, " +
-              "'count', 'group', " +
-              showMissingness + ")";
+          "plot.data::pie(" + DEFAULT_SINGLE_STREAM_NAME + ", map, " +
+              valueSpec + ", '" +
+              deprecatedShowMissingness + "')";
+              System.out.println(cmd);
```
asizemore, line 96: Leftover print statement?

asizemore (commented):
One big question - i didn't understand how/when ShowMissingnness vs ShowMissingnessNoAxes was applied?

schema/url/pass/plugin-barplot.raml
d-callan: yea its a good question. right now the only restriction i put was on where it was or wasnt currently working. but youre right it results in some very arbitrary decisions here... 
i guess if we make a dedicated map histo endpoint then i can apply the two pretty cleanly based on whether its for the map or not. and my current thinking is we should probably have a dedicated map histo anyway... so ill update it

src/main/java/org/veupathdb/service/eda/ds/plugin/abundance/AbundanceBoxplotPlugin.java
d-callan: lol i think my text editor played me a trick

asizemore:
Lovely! If i could double-approve, i would!

Issue #160, opened 2022-04-01
Backend: Need a map-histogram endpoint 


d-callan:
map-histogram will have different variables and associated constraints to the standard histo. some examples:

1. unlimited facets
2. allow 'allVariables' option for showMissingness (depends on veupathdb/plot.data#145)
3. no overlay variable
4. probably other differences i havent thought of yet

d-callan:
we could also consider clearer names for the map specific endpoints.. maybe pie -> 'map-categorical-markers', this new one -> 'map-continuous-markers' and map-marker -> 'map-marker-entity-counts'

d-callan:
related to veupathdb/edanewissues#217

d-callan:
to clarify there is i think only a single facet var but w unlimited number of values. another difference between this and the regular histo is i think the values returned by binSlider. since were functionally using this to discretize continuous data into a bar/pie plot w ~8 values as opposed to a proper histogram the binSlider will likely have to set a much lower max than 1000 bins.

d-callan:
so given we want different bin slider behavior to the standard histo, require support for a geolocation viewport specifically in this case and are considering supporting discretization of continuous variables for all plot classes in plot.data (thereby possibly reducing or removing the motivation to use the histo endpoint here)... im now inclined to incorporating the needed features into the existing pie endpoint and renaming that something map specific. @bobular @asizemore thoughts?

bobular:
That sounds wise.

Just to help me understand (and not directly related to maps): by "discretization of continuous variables for all plot classes" does that mean that a continuous variable could be chosen for the main variable of a barplot? Or just for overlays?

bobular:
Assuming just overlays, would the assumption be that 99% of overlay use cases would use the max number of bins? (8?)

The map marker distribution endpoint (née pieplot) would be a special case offering control over the number of bins?

d-callan:
How to do the binning is maybe a hard question. If the goal is to keep the legend consistent between mapveu map and stats charts then I think the stats charts need to be able to support user specified number of bins also. But I think maybe the bins need to be determined before removing incomplete cases or possibly even viewport adjustment to ensure they are binned over the same range in both places. That or the range to bin over needs to be provided explicitly, maybe using variable metadata. That second is probably best in general actually.

d-callan:
But the front end could do either if that's your question.. the eda can do whichever the eda ux group likes

PR #161, opened 2022-04-04
Map Histogram

d-callan:
resolves veupathdb/edadataservice#160
depends on veupathdb/plot.data#145 and veupathdb/edadataservice#150


d-callan:
see linked issue, changing plans some here.

PR #165, opened 2022-04-07
map-markers complete cases count

### LINKED ISSUE #163: Backend: map-markers endpoint needs to accept a viewport and return completeCasesAxesVars
in order to get the birds eye bar working, we need to know the count of relevant entity records that have geo data available based on the users filter criteria. currently the viewport is being added to the filters, making this impossible. Instead we want to map-markers endpoint to take a viewport prop and find completeCasesAxesVars pre-viewport-adjustment. (similar to how histo does for ex)

d-callan:
resolves #163


schema/url/visualizations.raml
```diff
@@ -36,6 +36,24 @@ types:
       xMax:
         type: string
 
+  LatitudeViewport:
```
asizemore, line 39: Suggestion - min and max don't seem specific to Latitutde. Maybe this could be a more general Viewport?

asizemore: May be reasonable to have xAxisViewport be a more general viewport if we go that direction, too?

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapPlugin.java
```diff
@@ -142,8 +173,11 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
         .getBytes()
       );
     }
-    // close array and enclosing object
-    out.write("]}".getBytes());
+    // add config and close
+    // bc the only var we take is geoVar these two values are the same
```
asizemore, line 177: I missed this - then how does it work when we have the x-axis variable? We want hair color as x axis on pies, say. Then do we not have a geo variable and hair color var?

asizemore (commented):
Just one real question from me!

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapPlugin.java
d-callan: multiple requests.. this gives the location of map markers and entity counts only. to get data for the pies there is a second request to the pieplot endpoint where the facet variable is the geoVar. i mean really to call these axes vars here hardly makes sense. im considering not reusing PlotConfig and just having the and explicit single 'completeCasesGeoVar'

schema/url/visualizations.raml
d-callan: i made the xaxis one the way i did in case we end up w stats charts w y axis viewports as well... but you may be right. we could have NumericViewport w min and max

schema/url/visualizations.raml
```diff
@@ -28,20 +28,14 @@ types:
       value?: number
       units?: BinUnits
 
-  XAxisViewport:
+  NumericViewport:
     additionalProperties: false
     properties:
       xMin:
```
asizemore, line 34: If we want to keep NumericViewport pretty general, should probably have properties min and max instead of xMin and xMax? 

I recognize my suggestion may end up causing a lot of work changing xMin to min and so on, here and in plot.data. Since this is a b57 thing I'm also happy to make a new ticket for the name change and we can do it when time allows if you'd prefer?

asizemore (approved):
I really like "completeCasesGeoVar". Makes it very clear!

I had one question idk if you feel like needs to be addressed asap or we can fix it after this PR gets merged.

schema/url/visualizations.raml
d-callan: Yea it needs fixing in a number of repos and didn't seem urgent so I was planning to wait. New issue seems fine, but it probably is actually three tickets? Here, plot.data and web-eda I'd imagine at the least

PR #184, opened 2022-05-05
continuous map markers

### LINKED ISSUE #160: Backend: Need a map-histogram endpoint

d-callan:
resolves #160 

NOTE: !! API changes relevant to the map
depends on veupathdb/plot.data#166


schema/url/pass/plugin-map-markers-overlay.raml
```diff
@@ -18,45 +18,48 @@ types:
         required: false
       xAxisVariable: 
         type: VariableSpec
-      # TODO rename to geoAggregateVariable after b57
-      facetVariable: 
-        type: VariableSpec[]
-        maxItems: 2
-        required: false
+      geoAggregateVariable: 
+        type: VariableSpec
```
asizemore, line 22: i dont follow - why did we change to VariableSpec (from VariableSpec[])?

schema/url/pass/plugin-map-markers-overlay.raml
```diff
@@ -18,45 +18,48 @@ types:
...
+        type: BinSpec
       viewport?:
         type: GeolocationViewport
 
-  PieplotConfig:
+  MapMarkersOverlayConfig:
     type: PlotConfig
     properties:
-      rankedValues:
+      rankedValues?:
```
asizemore, line 37: why did this become optional?

asizemore (commented):
Pumped to see the plot.data changes!

Please excuse my basic questions - it's been a while since i thought about the map!

schema/url/pass/plugin-map-markers-overlay.raml
d-callan: Should only take a single geohash at a time. It's a new type of strata var essentially. If we need facets in the map viz at some point well add them back then, but they're not currently spec'd.

schema/url/pass/plugin-map-markers-overlay.raml
d-callan: I figured if we're passing back binned continuous data maybe we didn't need it? It was added so the client would know what legend entries it needed iirc. Though I suppose it wouldn't hurt to have in either case if we just wanted to be consistent, or had some other use case.

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapMarkersOverlayPlugin.java
```diff
@@ -14,94 +14,109 @@
...
+        } else {
+          binWidth = binSpec.getValue() == null ? binWidth : "'" + binSpec.getValue().toString() + " " + binSpec.getUnits().toString().toLowerCase() + "'";
+        }
+        connection.voidEval("binWidth <- " + binWidth);
+      }
+
       String cmd =
-          "plot.data::pie(" + DEFAULT_SINGLE_STREAM_NAME + ", map, " +
-              valueSpec + ", viewport, '" +
+          "plot.data::mapMarkers(" + DEFAULT_SINGLE_STREAM_NAME + ", map, binWidth, " +
```
bobular, line 117: plot.data::mapMarkers is still pending a push to plot.data repo?

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapMarkersOverlayPlugin.java
```diff
@@ -14,94 +14,109 @@
...
       
+      String binReportValue = "NULL";
+      if (xVarType.equals("STRING")) {
+        // maybe i should have a warning here if they pass a BinSpec?
+        connection.voidEval("binWidth <- NULL");
+      } else {
+        BinSpec binSpec = spec.getBinSpec();
+        validateBinSpec(binSpec, xVarType);
+        binReportValue = binSpec.getType().getValue() != null ? binSpec.getType().getValue() : "binWidth";
+        String binWidth = connection.eval("numBinsToBinWidth(" + DEFAULT_SINGLE_STREAM_NAME + "[[map$variableId[map$plotRef == 'xAxisVariable']]], 8)").toString();
```
bobular, line 107: A lot of stuff to unpack here. By default we will calculate the binWidth in R with `numBinsToBinWidth` but I would need some help with the `DEFAULT_SINGLE_STREAM_NAME + "[[map$variableId[map$plotRef == 'xAxisVariable']]]` bit. I'm assuming a pass through the data is made?  Is that on the subset?

We should go ahead with default/hardcoded number of bins == 8, and let the user tweak via bin **width** controls, but I wonder in the future we should let the user choose the number of bins?

bobular (approved):
Looks good in terms of API :-)

schema/library.raml
```diff
@@ -7,101 +7,138 @@
 ################################################################################
 
 types:
-  BoxplotPostRequest:
-    type: VisualizationRequestBase
-    properties:
-      config: BoxplotSpec
-  BoxplotSpec:
+  BarplotData:
```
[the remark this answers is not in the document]
d-callan: yes thats a generated file. i typically ignore it. its essentially a concatenation of the files under schema/url

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapMarkersOverlayPlugin.java
d-callan: see the recent veupathdb/plot.data#166

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapMarkersOverlayPlugin.java
d-callan: DEFAULT_SINGLE_STREAM_NAME is a reference to the data table which has been read into the rserve process, and everything inside the square brackets is essentially asking for the column id which is associated w the xaxis variable. the data at this point does always reflect the subset, but once its read into the r process we dont really need to worry about 'passes' through a stream. it just sits in memory.

d-callan:
rankedValues will tell you what the legend entries should be. if say the user chose a variable w a dozen values itd return the top 7 in order and 'other'.

PR #186, opened 2022-05-06
Local deployment using docker-compose

ryanrdoherty:


dmgaldi:
Here's a PR that I merged for the import service. The main difference between it and the code that's here is the `SSH_AUTH_SOCK` environment variable. This can differ on Mac vs. other Linux distros.

VEuPathDB/service-user-dataset-import#21

PR #193, opened 2022-06-02
explicit bin ranges for map markers overlay cont vars

d-callan:
depends on veupathdb/plot.data#171


d-callan:
if were happy w this as a strategy generally, then maybe i should think about allowing histo and line to take either viewport OR bin range for the short term and then eventually remove viewport entirely?

PR #198, opened 2022-07-06
explicitly define overlay vars as not required

### LINKED ISSUE #197: Variable Constraints: Do not require overlay variables
@jernestmyers noted when investigating [this issue](https://github.com/VEuPathDB/web-eda/issues/1177) that the data service is declaring overlay variables as required. I think this is just an oversight as we didn't explicitly say they are _not_ required (as we did with facets).

To Do

- [ ] For all visualizations where it makes sense (think about map!) set `isRequired` to false for overlay vars.

asizemore:
Resolves #197 

For each visualization, checked that the overlay variable is specifically not required. Tested by running the local backend and I'm now seeing the following "isRequired: false" in the response as expected:

(from barplot)
```
"dataElementConstraints": [
						{
							"xAxisVariable": {
								"isRequired": true,
								"minNumVars": 1,
								"maxNumVars": 1,
								"maxNumValues": 10,
								"description": "Variable must have 10 or fewer unique values."
							},
							"facetVariable": {
								"isRequired": false,
								"minNumVars": 1,
								"maxNumVars": 2,
								"maxNumValues": 10,
								"description": "Variable(s) must have 10 or fewer unique values and be of the same or a parent entity of the Overlay variable."
							},
							"overlayVariable": {
								"isRequired": false,
								"minNumVars": 1,
								"maxNumVars": 1,
								"maxNumValues": 8,
								"description": "Variable must have 8 or fewer unique values and be of the same or a parent entity as the X-axis variable."
							}
						}
					],
```

To the kind reviewers - i think only one review is necessary, given the simplicity of the change and ability to test. Just asking both of you in case one happens to be free sooner!


asizemore:
courtesy tag @jernestmyers in case you're interested!

PR #201, opened 2022-07-28
ensure scatter x axis constraints are consistent across apps

### LINKED ISSUE #200: All scatterplots should accept dates, integers, and numbers
Make the var constraints of alpha div and ranked abundance scatterplots match the pass app scatter's constraints.

asizemore:
Resolves #200 

The alpha div and ranked abundance scatters do not currently allow integers. The pass app scatter does. In order to make all these scatters consistent, this PR updates the alpha div and ranked abundance scatter to have the same x axis constraints as the pass scatter.


asizemore:
@bobular i think just a look at the code for now is good!

bobular (approved):
Looks good to me.

Now matches xAxisVariable constraints in
https://github.com/VEuPathDB/EdaDataService/blob/494fc014f546f797cd2853bd41bcf9e2967db20e/src/main/java/org/veupathdb/service/eda/ds/plugin/pass/ScatterplotPlugin.java

Issue #179, opened 2022-04-14
Standardize capitalization on viz tools


SheenaTomko:
We have a mix of sentence and title case for each viz tool. These should all be the same. I have a slight preference towards sentence case. The fix should also be applied to viz tools that are not yet avialable (Density plot, Pie plot, etc.)
<img width="1000" alt="Screen Shot 2022-04-14 at 3 13 52 PM" src="https://user-images.githubusercontent.com/54597319/163468362-77467d86-3370-45bd-a93d-cf28ea6955d3.png">

asizemore:
I'll be doing some data service work soon, so i might as well tackle this ticket, too!

asizemore:
Closed with #210 
Deployed to feature mbio and qa clinepi

danicahelb:
looks good! 

![Image](https://user-images.githubusercontent.com/25450900/189185047-75f2cfc1-ee44-47c2-8e60-7d0a91a20bf8.png)

PR #210, opened 2022-09-07
Fix viz titles and order

### LINKED ISSUE #209: Visualizations: improve order of mbio vizs
seems to me like histogram should probably be listed first under distributions, barplot first under counts and proportions etc. i also noticed a few places where sometimes plot was lower case and other times upper like 'Line plot' and 'Bar Plot'..

asizemore:
Resolves #209 and #179 

All visualizations should now have titles in sentence case. I also adjusted the ordering of visualizations in the Distributions and Counts + Proportions apps. Other apps looked okay. Based on the example analyses, alpha div boxes and scatters are approximately equally useful.


.vscode/configurationCache.log
```diff
@@ -0,0 +1 @@
+{"buildTargets":["build/libs/service.jar","clean","compile","default","docker","gen-docs","gen-jaxrs","install-dev-env","jar","merge-raml","test"],"launchTargets":[],"customConfigurationProvider":{"workspaceBrowse":{"browsePath":[],"compilerArgs":[]},"fileIndex":[]}}
```
d-callan, line 1: something snuck in?

d-callan (approved):

.vscode/configurationCache.log
asizemore: fixed!

PR #214, opened 2022-10-11
Refactor computed variable metadata

d-callan:
again see https://github.com/VEuPathDB/web-eda/issues/1329 for bigger picture. here also trying to streamline variable metadata in various responses. depends on https://github.com/VEuPathDB/EdaCommon/pull/19


src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java
```diff
@@ -71,20 +71,24 @@ public class AppsMetadata {
           viz("scatterplot", new BetaDivScatterplotPlugin())),
       app("distributions", "Distributions", "null",
           "Plot simple distributions for any continuous variable, including metadata (e.g. age, height, etc.) or microbial assay results.",
-          List.of(MICROBIOME_PROJECT),
+          List.of(MICROBIOME_PROJECT, CLINEPI_PROJECT, ALLCLINEPI_PROJECT),
```
ryanrdoherty, line 74: I think AllClinEpiDB is no more; we can probably remove at these locations.

ryanrdoherty (approved):

src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java
d-callan: this was intended for a different branch, pushed erroneously as i forgot i had done this.. fixed now.

schema/url/plots.raml
```diff
@@ -307,9 +299,7 @@ types:
     properties:
       completeCasesAllVars: number
       completeCasesAxesVars: number
-      computedVariableMetadata:
-        type: ComputedVariableMetadata
-        required: false
+      variables: VariableMetadata[]
```
bobular, line 302: How would I know to look for the raml definition of `VariableMetadata` in EdaCommon?

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/HistogramPlugin.java
```diff
@@ -113,7 +113,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
           spec.getOverlayVariable(),
           util.getVariableSpecFromList(spec.getFacetVariable(), 0),
           util.getVariableSpecFromList(spec.getFacetVariable(), 1)));
-      connection.voidEval(getVoidEvalVarMetadataMap(DEFAULT_SINGLE_STREAM_NAME, varMap));
+      connection.voidEval(getVoidEvalVariableMetadataList(varMap));
```
bobular, line 116: This half R half Java code is fairly impenetrable.  It would help to have a comment after a line like this just saying `// variables <- ...` for example

bobular (approved):

PR #215, opened 2022-10-13
Reorganize clinepi apps

d-callan:
needed for https://github.com/VEuPathDB/web-eda/issues/1406


d-callan:
setting as draft since it seems a while before well get to this

Issue #213, opened 2022-09-12
Mosaic plot (2x2) has variable constraint issue with terms allowed for x-axis and y-axis


danicahelb:
We discussed the variable constraint issue with 2x2 Mosaic Plot viz tool in 9/12/2022 EDA UX meeting.

Currently, the table cannot be set up properly for many Exposures of interest
- Proper set up: Outcome on the X-axis and Exposure on the Y-axis
- Issue: outcomes are often under “Sample” category, while exposures are often under “Household”, “Participant”, or “Participant repeated measures” categories

@d-callan proposed possibly “recycling” other variable constraint fixes, like revising data dependency order to allow facet & overlay to be independent: https://github.com/VEuPathDB/EdaDataService/issues/190 (also possibly related to https://github.com/VEuPathDB/EdaDataService/issues/119)

Danielle, can you take a look and see what the best approach for fixing the variable constraint issue will be? thanks!

sweverschulman:
I went to run a 2x2 plot in India ICEMR Cohort and received the following error. 

![Image](https://user-images.githubusercontent.com/62298921/200023213-713ba547-4c00-4390-b80f-b3392f49d5e6.png)

`Unexpected backend type(s) encountered: Invalid value null' supplied to relativerisk (type: number) of O (type: Partial<{ oddsratio:
(number | null), pvalue: (number | string), orInterval: string, rrinterval: string, relativerisk: number, facetVariableDetails: (I({ entityld:
string, variableld: string ) & ( value: string })I | [(( entityld: string, variableld: string ) & ( value: string }), (( entityld: string, variableld:
string } & { value: string })]) }>) of statsTable (type: Array<Partial<{ oddsratio: (number | null), pvalue: (number | string), orInterval:
string, interval: string, relativerisk: number, facetVariableDetails: ([({ entityld: string, variableld: string } & { value: string })] I [(f
entityld: string, variableld: string } & { value: string }), ({ entityld: string, variableld: string } & { value: string })!) }>>) of 1 (type:
Partial<{ statsTable: Array<Partial<{ oddsratio: (number | null), pvalue: (number | string), orInterval: string, rrinterval: string,
relativerisk: number, facetVariableDetails: ([({ entityld: string, variableld: string } & { value: string })I | [({ entityld: string, variableld:
string } & { value: string }), (( entityld: string, variableld: string ) & { value: string })I) }>>, completeCasesTable: Array<Partial<{
[...]

Issue #190, opened 2022-05-27
Backend: improve dataElementDependencyOrder to allow facet and overlay to be mutually independent


d-callan:
came up during a conversation w DSR where he was trying to approximately replicate a plot from a data provider and wasnt able to.

currently we can support a dependency order of visual elements defined something like:
'yaxis' -> 'xaxis' -> 'overlay' -> 'facet'

and wed like to be able to do something like:
'yaxis' -> 'xaxis' -> ['overlay','facet']

essentially stating that both the overlay and facet strata vars need to simply be the same or a parent entity of the xaxis, rather than the facet var being required as the same or parent of the overlay.

dmfalke:
Are you thinking the property would have a nested array, e.g.: `['yaxis', 'xaxis', ['overlay', 'facet']]`?

d-callan:
Tbh I hadn't thought much of anything yet.. do you have a preference?

dmfalke:
I think a nested array, as above, works.

sweverschulman:
@dmfalke and @d-callan - to confirm functionality. Facet and overly do not need to have the same entity level. Facet and overly does need to be the same entity level or parent of the axis?

d-callan:
Worth spending a bit of extra time in the mosaics, the same logic/equivalency applies to the axes now. Should be sure they work as desired.

sweverschulman:
I am messing around with RxC plot in Wash B Kenya. If I choose x-axis of “died” (participant) and y-axis cluster study arm. I am not allowed to choose ‘sex’ which has same parent as ‘died’. So I am a little confused at while it isn't allowing faceting

d-callan:
@bobular i think somewhere i said that for the axes when equivalent, as in mosaic, to use the most ancestral entity between the two for the variable constraints dependency order.. but really we should be using the child most entity as the examples from steph and danica above illustrate. think this needs fixed front end. sry for the confusion.

bobular:
> @bobular i think somewhere i said that for the axes when equivalent, as in mosaic, to use the most ancestral entity between the two for the variable constraints dependency order.. but really we should be using the child most entity as the examples from steph and danica above illustrate. think this needs fixed front end. sry for the confusion.

The strange thing is that I did implement it as childmost. (Though I forgot to rename one of the functions.)  But I will investigate deeper.

bobular:
Ah, that was for determining the output entity.

In the main constraints handling I think I was still using the most-ancestral logic.

[...]

Issue #119, opened 2021-10-13
Visualizations: revisit variable constraints


danicahelb:
We need to decide on appropriate variable constraints for visualizations. 

See https://github.com/VEuPathDB/EdaDataService/issues/106

d-callan:
for reference, i believe the specific proposal was to consider pros and cons of replacing the dependency order of variable constraints (ex: yaxis -> xaxis -> overlay -> facet) w a single dependency for the entity space of the plot on a particular variable. in that case you might just specify 'yaxis' was the variable that set the entity space of the plot, and ALL other variables had to be of the same or a parent entity of the yaxis.

danicahelb:
Issue with variable constraints for 2x2 contingency tables (mosaic viz tool)

Currently, the table cannot be set up properly (Outcome on the X-axis and Exposure on the Y-axis) for many Exposures of interest. 

Issue: outcomes are often under “Sample” category, while exposures are often under “Household”, “Participant”, or “Participant repeated measures” categories

danicahelb:
see: https://github.com/VEuPathDB/web-eda/issues/575

using the WASH-b Kenya study, I made a box plot of LAZ (participant repeated measure entity) for each study arm (community entity). I'd like to overlay this by either "study timepoint" or "diarrhea in last 7 days" (both under participant repeated measure entity) but I can only select overlay variables in the community entity

![image](https://user-images.githubusercontent.com/25450900/135361826-6c093b9c-7a4f-4820-9663-6cf6f566999e.png)

@d-callan stated:
> currently this is by design. we've implemented variable constraints in boxplot so that the strata vars have to be of the same or parent entity of the xaxis term. it seems like what you're proposing is that rather than have some dependency order like yaxis -> xaxis -> overlay -> facet, we might have only the var that sets the entity space of the plot specified (in this case yaxis) and ALL other vars must simply be of the same or parent entity as that one. seems like we should carefully consider pros and cons of that.

danicahelb:
I'm reopening this ticket. 

Let's collect use cases that the current system does not allow and we can revisit the variable constraints in the future once we have more evidence of issues.

danicahelb:
see: https://github.com/VEuPathDB/EdaDataService/issues/213

WASHb Bangladesh

the outcome i want to investigate is "Diarrhea case during the last 7 days, by caregiver report"
- binary values: yes / no
- participant repeated measure

the exposure i want to investigate is "Chlorine detected in stored drinking water"
[...]

PR #216, opened 2022-10-13
update variable constraints dependency order

d-callan:
this should get us things like [yaxis], [xaxis], [overlay, facet] 
resolves https://github.com/VEuPathDB/EdaDataService/issues/190
depends on https://github.com/VEuPathDB/EdaCommon/pull/20

may also resolve https://github.com/VEuPathDB/EdaDataService/issues/213
and obviate the need for https://github.com/VEuPathDB/EdaDataService/issues/119


src/main/java/org/veupathdb/service/eda/ds/plugin/abundance/AbundanceBoxplotPlugin.java
```diff
@@ -55,7 +55,7 @@ protected Class<AbundanceComputeConfig> getComputeSpecClass() {
   @Override
   public ConstraintSpec getConstraintSpec() {
     return new ConstraintSpec()
-      .dependencyOrder("overlayVariable", "facetVariable")
+      .dependencyOrder(new String[] {"overlayVariable", "facetVariable"})
```
ryanrdoherty, line 58: @d-callan If you're willing to go back to all these, you can get rid of some boilerplate like this:
```
  String[][] blah = new String[][] {
    {"a","b"}, {"c","d"}
  };
```

ryanrdoherty (approved):

d-callan:
@bobular maybe confirm the way ive set up the new dependency order makes sense to you as well? and works for front end dev? 

@ryanrdoherty how does merging this work given the edacommon dependency? i imagine i should include a version update for the dependency somewhere in this pr?

ryanrdoherty:
@d-callan The steps would be to merge the EdaCommon PR to main and then release it, then update this PR with the new version.  But if you do that before the client is ready, we won't be able to add new stuff to main (e.g. the compute client stuff I'm working on).  Is the client stuff ready to go out too?

d-callan:
@ryanrdoherty maybe I'm just being dense but why can't we add stuff to main in that case

ryanrdoherty:
We can but it will break dev sites, right?

bobular:
I'm just guessing after grepping some files, but would it be possible to add a temporary tag to the corresponding branch in https://github.com/VEuPathDB/EdaCommon/pull/20 - e.g. something like `v9.3.0pre`

Then update `build.gradle.kts` in this repo?
```
val edaCommon     = "9.3.0pre"         // EDA Common version
```

Then I might be able to get the `eda-data` service running locally via docker-compose magic?

bobular:
Hi @d-callan 

I think there could be an issue with the array of arrays representation.

When we are talking about stratification variables, we want this:

`[["xAxisVariable"],["overlayVariable","facetVariable"]]`

to mean "overlay and facet variables are both constrained by xAxisVariable, but not by each other"

This is how I initially interpreted it.

But when I did the same for contingency tables, which have this dependency order:

`[["yAxisVariable","xAxisVariable"],["facetVariable"]]`

I constrained y and x variables only to be on the same branch as each other, but not depend on each other in any specific order.

So this would allow
```
y-axis: participant sex
x-axis: household water availability
```
which of course the current live site does using the old approach.

and also the other way around:
```
y-axis: household water availability
x-axis: participant sex
```
But the back end doesn't allow this. It complains about the variables being the wrong way round in the hierarchy.

Are we going to change the back end to allow both "directions" (x-y, y-x) for contingency tables?

If we don't, then I need a way to know that
``[["xAxisVariable"],["overlayVariable","facetVariable"]]`` doesn't mean `facetVariable` depends on `overlayVariable`
but in `[["yAxisVariable","xAxisVariable"],["facetVariable"]]` it DOES mean that `xAxisVariable` depends on `yAxisVariable`.

An alternative would be to use the old-style dependencies for cont tables, e.g.
`[["yAxisVariable"],["xAxisVariable"],["facetVariable"]]`

d-callan:
I think the backed needs updating too.

bobular:
OK, cool. I've just implemented it back to the my original interpretation and it seems to be working well for stratification. :smile:

bobular (approved):
All good. Don't forget to upgrade the "9.3.0pre" to, I guess, "9.3.0" when the other PR is merged.

Issue #235, opened 2023-02-01
update 2x2 plugin to support new mosaic plot class api


d-callan:
depends on https://github.com/VEuPathDB/plot.data/issues/202 and https://github.com/VEuPathDB/plot.data/issues/203

PR #236, opened 2023-02-01
2x2 stats table

d-callan:
resolves https://github.com/VEuPathDB/EdaDataService/issues/235
depends on https://github.com/VEuPathDB/plot.data/pull/204 and https://github.com/VEuPathDB/veupathUtils/pull/11


schema/url/pass/plugin-twobytwo.raml
```diff
@@ -3,23 +3,52 @@
...
+      config: TwoByTwoSpec
+
+  # TODO once we get the other stats tables on this path we can add df, configuration notes, error msg etc
+  Statistic:
+    additionalProperties: false
+    properties:
+      value: number
+      confidenceInterval?: string # (min - max)
+      confidenceLevel?: number
+      pvalue: string # number to two signifs or string '<0.0001'
```
asizemore, line 24: Or NA?

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/TwoByTwoPlugin.java
```diff
@@ -104,7 +106,11 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
           util.getVariableSpecFromList(spec.getFacetVariable(), 0),
           util.getVariableSpecFromList(spec.getFacetVariable(), 1)));
       connection.voidEval(getVoidEvalVariableMetadataList(varMap));
-      String cmd = "plot.data::mosaic(" + DEFAULT_SINGLE_STREAM_NAME + ", variables, 'bothRatios', '" + deprecatedShowMissingness + "')";
+      String cmd = "plot.data::mosaic(data=" + DEFAULT_SINGLE_STREAM_NAME + ", variables=variables, statistic='all', columnReferenceValue=" + 
+                                          colRefValue + ", rowReferenceValue=" + 
+                                          rowRefValue + ",'" + 
+                                          deprecatedShowMissingness + "')";
+                                          System.err.println(cmd);
```
asizemore, line 113: Can be removed?

asizemore (approved):
Really digging the Statistic type. It makes TwoByTwoStatsTable so much cleaner

schema/url/pass/plugin-twobytwo.raml
d-callan: That's a good question. Right now an NA in plot.data gets written to Json as a null. So it's a question of whether we prefer that or a string 'NA' for these I guess..

asizemore: i would think null would be easier to handle in the frontend, instead of checking for if that string = NA. But perhaps we already do this elsewhere and have a better handling of it than i know.

PR #240, opened 2023-02-07
Filter aware variable metadata

### LINKED ISSUE #239: Standalone map: plugin to find bin ranges for continuous variables based on subset
As discussed here: https://veupathdb.atlassian.net/wiki/spaces/WG/pages/181796873/2023-02-02+Map+UX 

Default behaviour could continue to be equal spaced bins using global min & max

However, we want the binning to use the subset min and max (ignoring viewport) which will require a separate back end request to calculate this.

Then also consider quantile and SD-based binning.

Approx task list

- [ ] new `continuous-var-bin-spec` (feel free to rename) end point, calculates min, max and equal binWidth 
- [ ] add quantile-based bin mode for `continuous-var-bin-spec` (again this needs to be calculated on the subset, not the viewport-ed subset) - this presumably returns a list of binStarts and ends or similar
- [ ] ~~allow `map-markers-overlay` to handle quantile binSpec~~
- [ ] similar approach for SD-based binning (but probably for Phase X)

d-callan:
There have been a number of conversations about this recently. This work is needed in the immediate term for the standalone map application, and very likely will be needed for the EDA for clinepi and mbio as well in the near future. This PR merely represents a proposed API for us to discuss at this point. If people like it, maybe itll also become a place to start dev in some earnest.

The need is essentially to be able to find arbitrary variable metadata for any particular variable, but to also have that metadata be sensitive to any filters that have been applied to that variable. I have included here endpoints for bin ranges and medians, since I anticipate those being the first two use cases.

resolves #239

depends on https://github.com/VEuPathDB/veupathUtils/pull/12


d-callan:
just for clarity, i think overall were saying the client needs to:
1. consider filters on variables when it determines if those variables meet the specified constraints for a variable selection in a viz
2. consider filters on variables when determining what values/ ranges to show for axes and gradient overlays for those variables. when doing this it should still respect annotated display ranges. so for cont vars i think the filter min/max should replace the observed min/max of that study in the logic for finding ranges for axes and gradients, but the underlying logic for what ranges to show for axes and gradients should stay the same.
3. ask the backend for specific filter sensitive metadata when it needs to do things like produce binned gradient overlays, then somehow hold on to that knowledge so it can pass that metadata to the appropriate viz plugin

d-callan:
@ryanrdoherty @dmfalke ive reworked this some to adopt daves suggestion of a single configurable endpoint (not 100% sure i got the raml right but can figure that later). do we like this better?

also, i wonder what we think about my departure from VisualizationRequestBase/ not allowing all filters to be passed to this endpoint in favor of just asking for the one that pertains to the particular variable we need metadata for?

ryanrdoherty:
> @ryanrdoherty @dmfalke ive reworked this some to adopt daves suggestion of a single configurable endpoint (not 100% sure i got the raml right but can figure that later). do we like this better?
> 
> also, i wonder what we think about my departure from VisualizationRequestBase/ not allowing all filters to be passed to this endpoint in favor of just asking for the one that pertains to the particular variable we need metadata for?

I think I prefer we go with our existing pattern.  We could have a separate app that contains this, however, to indicate that its responses are not meant to directly feed into visualizations on the client.  This new app would not have a compute and could have an empty project list (telling the client not to show it anywhere), but would still be callable by other viz plugins.

d-callan:
Knowing the new plugin will be called by other plugins rather than the client does make me feel a little better about reusing VisualizationRequestBase. But I still feel odd and a bit unhappy calling the new plugin a 'visualization' or talking about it belonging to an 'app'. That language and those concepts don't make sense in the same way for this case.

Maybe that could be improved though simply by renaming VisualizationRequestBase to RequestBase?

ryanrdoherty:
Yeah, I understand.  You know, the way this is rigged up, we don't even have to put the plugin in the metadata or (I think) in api.raml.  However, you do still need to write the raml types.  We could have a pseudo "app" under the plugins directory called "internal" that contains these- _if_ they are only called internally.

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+    if (_cachedResponse != null) LOG.info("Found cached result.");
+
+    return _cachedResponse != null ? Collections.emptyList() : ListBuilder.asList(
+      new StreamSpec(DEFAULT_SINGLE_STREAM_NAME, pluginSpec.getVariable().getEntityId())
+        .addVar(pluginSpec.getVariable()));
+  }
+
+  @Override
+  protected void writeResults(OutputStream out, Map<String, InputStream> dataStreams) throws IOException {
+    CategoricalCountsSpec spec = getPluginSpec();
```
ryanrdoherty, line 90: You're never adding the result to the cache here.  The condition on line 83 says: "if we found the result in the cache, we don't need to request a stream", but then in writeResults() you read the stream anyway, so that would give you a NullPointerException on 93.  The trick is to check again if _cachedReponse is null (that's why it and _cacheKey are fields, not local vars), and return if so.  Otherwise you have to call RESULT_CACHE.getValue() where you essentially tell the cache how to create the value to be put into the cache.  This API is so the creation+insert+return happens atomically and is threadsafe.  See [the impl here](https://github.com/VEuPathDB/EdaDataService/blob/master/src/main/java/org/veupathdb/service/eda/ds/plugin/sample/RecordCountPlugin.java) for an example.

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+
+    // create scanner and line parser
+    InputStreamReader isReader = new InputStreamReader(new BufferedInputStream(dataStreams.get(DEFAULT_SINGLE_STREAM_NAME)));
+    BufferedReader reader = new BufferedReader(isReader);
+    DelimitedDataParser parser = new DelimitedDataParser(reader.readLine(), TAB, true);
+
+    // establish column header indexes
+    Function<VariableSpec,Integer> indexOf = var -> parser.indexOfColumn(getUtil().toColNameOrEmpty(var)).orElseThrow();
```
ryanrdoherty, line 98: This seems like overkill here since `indexOf` is only used once.  Why not just call parser.indexOfColumn with spec.getVariable()?

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+      .pattern()
+        .element("variable")
+          .shapes(APIVariableDataShape.CATEGORICAL, APIVariableDataShape.ORDINAL)
+          .description("Variable must be categorical.")
+      .done();
+  }
+  
+  @Override
+  protected void validateVisualizationSpec(CategoricalCountsSpec pluginSpec) throws ValidationException {
+    validateInputs(new DataElementSet()
```
ryanrdoherty, line 69: Eyeroll.  We should probably implement this some day.

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+      String[] row = parser.parseLineToArray(nextLine);
+
+      categoryCounts.putIfAbsent(row[varIndex], new MutableInt());
+      categoryCounts.get(row[varIndex]).increment();
+
+      nextLine = reader.readLine();
+    }
+
+    // begin output object and single property containing array of elements
+    out.write("{\"categories\":[".getBytes(StandardCharsets.UTF_8));
```
ryanrdoherty, line 115: Looks like this just creates a histogram of a categorical var for a given subset.  If that's all we need, I could probably just expose a call to the subsetting service's distribution endpoint (with subset of your choice).  You'd have to parse that endpoint's API but it should be a lot quicker than this.  Only downside I can think of is that you couldn't use a derived var (for now).  Is there a reason this wouldn't be better?

ryanrdoherty: p.s. Dan Galdi is beginning to think about changing over the count and distribution endpoints to use his much faster file-based subsetting system.  Probably still not as fast as a cached value, but probably pretty quick.

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+    boolean first = true;
+    for (String key : categoryCounts.keySet()) {
+      // write commas between elements
+      if (first) first = false; else out.write(',');
+      out.write(new JSONObject()
```
ryanrdoherty, line 120: I would say the only reason to do this kind of streaming is if you're taking values from a stream and don't want to store the entire JSON result in memory.  But since you're already storing the entire result in the HashMap above, it's probably no less safe to simply write:
```
out.write(new JSONObject().put("categories",
   categoryCounts.entrySet().stream()
      .map(e -> new JSONObject().put("binLabel", e.getKey()).put("value", e.getValue()))
      .toList()
).toString().getBytes(StandardCharsets.UTF_8));
```

ryanrdoherty: Not even sure you need the StandardCharsets.UTF_8 since that is probably the default.

src/main/java/org/veupathdb/service/eda/ds/plugin/FilteredVariableMetadataPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+    if (_cachedResponse != null) LOG.info("Found cached result.");
+
+    return _cachedResponse != null ? Collections.emptyList() : ListBuilder.asList(
+      new StreamSpec(DEFAULT_SINGLE_STREAM_NAME, pluginSpec.getVariable().getEntityId())
+        .addVar(pluginSpec.getVariable()));
+  }
+
+  @Override
+  protected void writeResults(OutputStream out, Map<String, InputStream> dataStreams) throws IOException {
+    PluginUtil util = getUtil();
```
ryanrdoherty, line 90: Same comment as above about cache management

src/main/java/org/veupathdb/service/eda/ds/plugin/FilteredVariableMetadataPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+    return ContinuousVariableMetadataPostRequest.class;
+  }
+
+  @Override
+  protected Class<ContinuousVariableMetadataSpec> getVisualizationSpecClass() {
+    return ContinuousVariableMetadataSpec.class;
+  }
+
+  @Override
+  public ConstraintSpec getConstraintSpec() {
```
ryanrdoherty, line 57: Same comment as above about getting this info into the hands of the client

src/main/java/org/veupathdb/service/eda/ds/plugin/FilteredVariableMetadataPlugin.java
```diff
@@ -0,0 +1,128 @@
...
+    return _cachedResponse != null ? Collections.emptyList() : ListBuilder.asList(
+      new StreamSpec(DEFAULT_SINGLE_STREAM_NAME, pluginSpec.getVariable().getEntityId())
+        .addVar(pluginSpec.getVariable()));
+  }
+
+  @Override
+  protected void writeResults(OutputStream out, Map<String, InputStream> dataStreams) throws IOException {
+    PluginUtil util = getUtil();
+    ContinuousVariableMetadataSpec spec = getPluginSpec();
+    var jsonWrapper = new Object(){ String value = "{"; };
```
ryanrdoherty, line 92: Probably better to use a StringBuilder here with calls to append() below, but could also just use a JSONObject and build from there, converting to String at the end.

ryanrdoherty (changes requested):
Gave it a shot.  Main thing before this gets added is moving the actual endpoints around.  It's not great to add arbitrary endpoints here because we want a well defined (and not too large) prefix set for each service in the stack for traefik routing considerations.  But also what I mentioned in comments.

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
```diff
@@ -0,0 +1,140 @@
...
+    return CategoricalCountsPostRequest.class;
+  }
+
+  @Override
+  protected Class<CategoricalCountsSpec> getVisualizationSpecClass() {
+    return CategoricalCountsSpec.class;
+  }
+
+  @Override
+  public ConstraintSpec getConstraintSpec() {
```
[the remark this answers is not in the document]
d-callan: i thought about this some when i was drafting this.. and im not sure we do need to send these to the client? i mean they arent constraints in the sense weve meant up to this point since this isnt a viz where were trying to constrain some user selection.

src/main/java/org/veupathdb/service/eda/ds/plugin/CategoricalCountsPlugin.java
d-callan: i think i like this idea. curious if @bobular sees any reason to not do this?

src/main/java/org/veupathdb/service/eda/ds/plugin/filteredmetadata/ContinuousVariablePlugin.java
```diff
@@ -0,0 +1,142 @@
...
+    return _cachedResponse != null ? Collections.emptyList() : ListBuilder.asList(
+      new StreamSpec(DEFAULT_SINGLE_STREAM_NAME, pluginSpec.getVariable().getEntityId())
+        .addVar(pluginSpec.getVariable()));
+  }
+
+  @Override
+  protected void writeResults(OutputStream out, Map<String, InputStream> dataStreams) throws IOException {
+    if (_cachedResponse == null) {
+      LOG.info("Result not in cache; calculating..");
+      StringBuilder json = new StringBuilder("{");
```
ryanrdoherty, line 93: Can move this declaration inside the value producer lambda.

ryanrdoherty: Also can move the LOG line inside because we don't know for sure if we are calculating it until the function runs.

api.raml
```diff
@@ -26,6 +26,18 @@ securitySchemes:
 # Data Service Specific Endpoints
 ####################################
 
+/filter-aware-metadata/continuous-variable:
```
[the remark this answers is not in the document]
ryanrdoherty: Side note, we don't actually even have to expose this as an endpoint if we don't want to (would still need to be wired up in the service class anyway), since it can be called directly by the dependent plugins.  The only thing we actually lose is testability.  Not saying we shouldn't expose, just that we don't have to.

ryanrdoherty (approved):

schema/url/filtered-variable-metadata.raml
```diff
@@ -0,0 +1,46 @@
...
+    properties:
+      variable: VariableSpec
+      metadata:
+        # is this valid? im guessing how wed do this..
+        type: array
+        items:
+          enum: ['binRanges', 'median']
+
+  # TODO should the BinRange and BinRangeWithValue live somewhere else?
+  BinRange:
```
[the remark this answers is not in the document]
d-callan: i think i need optional bin start, end and value. can we change HistogramBin to accomodate that?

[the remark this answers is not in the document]
ryanrdoherty: Yes, we could set it to be optional or use special strings like "INF" and "-INF".  Though for dates should probably be "ETR" and "-ETR".  That was an attempt at a joke.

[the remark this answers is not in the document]
d-callan: if youre happy to make them optional, thatd be my preference.

schema/url/filtered-variable-metadata.raml
```diff
@@ -0,0 +1,46 @@
...
+
+types:
+  
+  ContinuousVariableMetadataPostRequest:
+    additionalProperties: false
+    type: VisualizationRequestBase
+    properties:
+      config: ContinuousVariableMetadataSpec
+
+  ContinuousVariableMetadataSpec:
```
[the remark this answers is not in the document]
d-callan: idk if the subset will remain consistent enough to make calculating them all really worthwhile. ill plan to update the key

api.raml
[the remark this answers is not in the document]
ryanrdoherty: Ok, sounds good.  I'm fine with this endpoint name.

PR #241, opened 2023-02-08
Demonstrate a simple visualization result cache

ryanrdoherty:
This PR demonstrates how we can cache results of the future stats endpoint for use by other plugins or by the web client.


d-callan:
This is really great 👍
Only thing is I don't think that the new endpoint needs to know about all the filters, just the one pertaining to the requested variable. So I'd guess we cache based on the plugin config rather than the filters + ID.

d-callan (approved):

PR #242, opened 2023-02-08
Add infrastructure and example allowing plugins to call each other

ryanrdoherty:


d-callan:
I like this except for one thing: I want different filters for each of the two plugins. I think the calling plugin needs to be able to choose which of it's filters to pass along. For the planned bin ranges for ex, we want to be able to apply only the filter that pertains to the variable we are requesting bins for. 

As for why, ill make up a trivial example: i explicitly filter against people below a certain height, and incidentally remove whole geographic locations, say villages or something. the resulting plot should provide explicit 0 values for those villages since we didnt filter against them. its good to know that the filters we did apply removed them, in both the subset and viz tabs. plus, tbh idk how wed ever do constraints in the viz tab in a performant way if every variables metadata needed to be found again every time any filter changed.

PR #243, opened 2023-02-09
Standalone map

### LINKED ISSUE #237: Standalone map: create new "pass-through app" for map marker and viz endpoints
Although the regular EDA and Map viz plugins will share a lot of code, it makes sense to separate their APIs so they can be customised cleanly as needed.

For example, we can get rid of the facet variable for all the map viz's

Also birds eye and completeness data may not be needed for the SAM

### LINKED ISSUE #238: Standalone map: categorical overlay, update API for customisable N<=7 + other
The new-improved handling of categoricals including high-cardinality vars as discussed [here](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/181796873/2023-02-02+Map+UX) is to
- if the variable has >8 values, continue to overlay most-populous-7-in-viewport + pooled Other as the **default** (did we actually agree this?)
- if a list of up to 7 values are provided in the request, overlay by these + pooled Other, even if the variable has <=8 values

Approx task list
- [ ] handle POST payload in EdaDataService
- [ ] implement necessary R functions
- [ ] deploy to `map-markers-overlay` endpoint
- [ ] deploy to `lineplot` endpoint (as highest priority of the visualizations)
- [ ] deploy to other vizs

d-callan:
some refactoring for the standalone map now being its own application, and the initial draft of the markers and floating charts plugins.

resolves #237 
resolves #238

depends on https://github.com/VEuPathDB/EdaDataService/pull/240
depends on https://github.com/VEuPathDB/plot.data/issues/209
depends on https://github.com/VEuPathDB/plot.data/issues/206
depends on https://github.com/VEuPathDB/plot.data/issues/205
depends on https://github.com/VEuPathDB/plot.data/issues/215

related to https://github.com/VEuPathDB/plot.data/issues/200
related to https://github.com/VEuPathDB/plot.data/issues/210


schema/url/standalone-map/plugin-floating-lineplot.raml
```diff
@@ -0,0 +1,55 @@
...
+  FloatingLineplotSpec:
+    additionalProperties: false
+    properties:
+      outputEntityId:
+        type: string
+      xAxisVariable: 
+        type: VariableSpec
+      yAxisVariable: 
+        type: VariableSpec
+      yAxisNumeratorValues:
```
bobular, line 20: There's potentially another UI real estate challenge with the extended lineplot variables UI, but I think for now we keep this in and make space for it somehow.

schema/url/standalone-map/plugin-floating-lineplot.raml
```diff
@@ -0,0 +1,55 @@
...
+        type: StringBoolean
+      viewport:
+        type: NumericViewport
+        required: false
+
+  FloatingLineplot:
+    additionalProperties: false
+    properties:
+      data:
+        type: LineplotData[]
```
bobular, line 48: Are we OK using `LineplotData` as the response type, even though it will never contain (the optional) `facetVariableDetails`? I think so, but just wanted to raise this.

schema/url/standalone-map/plugin-floating-scatterplot.raml
```diff
@@ -0,0 +1,44 @@
...
+      config: FloatingScatterplotSpec
+
+  FloatingScatterplotSpec:
+    additionalProperties: false
+    properties:
+      outputEntityId:
+        type: string
+      valueSpec:
+        type: string
+        enum: ['raw', 'smoothedMeanWithRaw', 'bestFitLineWithRaw']
```
bobular, line 18: Just making a note to myself that there are potential performance issues with respect to moving around the map

1. many unaggregated points in the scatter plot (network, client rendering) - this goes against the whole ethos of server-side aggregation
2. line fitting is a kind of aggregation and avoids network+client issues but is more CPU intensive than the string-based marker geo-aggregation

So perhaps we should "ban" all non-performant visualizations for the SAM?

But that seems a bit mean.

More aggressive client-side debouncing for CPU-intensive viz types maybe?

bobular: And maybe only allow aggregating vizs (like the fitted lines without the raw points)?

bobular (commented):
This looks great.  I only looked through the schema for this review. Hopefully a second Java review will follow... :smile:

src/main/java/org/veupathdb/service/eda/ds/plugin/standalone-map/FloatingBarplotPlugin.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @Override
+  public ConstraintSpec getConstraintSpec() {
+    return new ConstraintSpec()
+      .dependencyOrder(List.of("xAxisVariable"), List.of("overlayVariable"))
+      .pattern()
+        .element("xAxisVariable")
+          .maxValues(10)
+          .description("Variable must have 10 or fewer unique values and be of the same or a child entity as the variable the map markers are painted with.")
```
bobular, line 63: Ah, I see. Nice!

src/main/java/org/veupathdb/service/eda/ds/plugin/standalone-map/StandaloneMapMarkersPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+        .put("avgLon", avgLatLon.getLongitude())
+        .put("minLat", data.minLat)
+        .put("minLon", data.minLon)
+        .put("maxLat", data.maxLat)
+        .put("maxLon", data.maxLon)
+        .toString()
+        .getBytes(StandardCharsets.UTF_8)
+      );
+    }
+    // add config and close
```
bobular, line 161: So is the plan to use pure Java for the overlay summary (categorical "barchart/donut" or continuous histogram)?

If so, I'd suggest sticking with R for now, while we build the client.

bobular (commented):
No major comments on the code. I still see a few mentions of facetVariable but I assume this is still WIP.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalone-map/StandaloneMapMarkersPlugin.java
d-callan: the current plan is to have the quantiles/ bin ranges/ filter-aware-var-metadata endpoint do its work in R, and this one in Java. The map plugin that belongs to the pass app will keep doing its work in R. i assumed youd keep building the client from that endpoint, and switch to this one when i finally have it working. i know that means an api change, but this wont be ready for a while yet probably. theres a good amount of work to do in R as well.

schema/url/standalone-map/plugin-standalone-map-markers.raml
```diff
@@ -0,0 +1,70 @@
...
+      geoAggregateVariable:
+        type: VariableSpec
+      longitudeVariable:
+        type: VariableSpec
+      latitudeVariable:
+        type: VariableSpec
+      overlayVariable?:
+        type: VariableSpec
+      overlayValues?:
+        type: BinRange[] # categoricals can just ignore bin start and end
```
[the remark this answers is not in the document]
d-callan: yea. maybe i should call it something other than BinRange? and i suppose the plugin will have to do some manual validation. if no bin start and ends are specified, and we only receive labels, it should probably complain if it gets more than 8 values. and it should maybe also complain if it gets more than 7 but sees 2+ additional unrequested values?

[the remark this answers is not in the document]
bobular: Oh yeah, that's simpler and already there!

schema/url/standalone-map/plugin-standalone-map-markers.raml
```diff
@@ -0,0 +1,70 @@
...
+      # if we choose to support things like means and medians for cont vars this may need to be a custom enum
```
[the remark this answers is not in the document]
bobular: Oops yes, that's simpler, very clear, and already there! I don't think 'binCounts' etc are needed.

schema/url/standalone-map/plugin-standalone-map-markers.raml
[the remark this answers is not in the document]
bobular: I was going to suggest for simplicity that the client always provides <=7 values. But the more I thought about it, the more complicated it gets, and the back end sanity checks like you describe would still be needed anyway.

schema/url/standalone-map/plugin-floating-lineplot.raml
bobular: Thanks, but I think the custom fractions for categoricals are probably going to be quite useful. At least I can't think of a reason to get rid of them at this stage. We'll find some screen space somehow.

schema/url/standalone-map/plugin-floating-scatterplot.raml
bobular: Yes, I understand. Let's limit the number of points, or disable scatter completely, if performance is an issue.

bobular:
Nice. I was just wondering about these variable/help documentation strings:

> Variable must be ordinal, a number, or a date and be the same or a child entity as the variable the map markers are painted with.

This assumes that the map markers are always painted with an overlay variable, which is something I've said we tentatively want to do and we can enforce via the UI (using the first featured variable at startup and not allowing the user to cancel the overlay variable).

Does this mean that the floating viz overlay variables should be `required(true)`?

We have a UX decision to make about what to do with changing the overlay variable when there are floating vizs already configured.

For the alpha FSM we allowed unconstrained selection of the overlay, and disabled the overlay in the floating viz if it became incompatible after a user-selection. Another option would be to grey-out/disable the whole floating viz until either the user reverts the overlay variable selection, or less likely, changes the main variables for that viz. I think floating viz should be allowed to have no overlay though, so the disabling behaviour described above probably still works.

Something to discuss on Thursday!

d-callan:
i can update the help text to make it clearer the overlay is optional.

as for ux, i agree the overlay selection should be unconstrained. but i wonder if we cant find some middle ground between gray plots and disabled overlays.. maybe the plot panel becomes gray until the user either actively changes the axes vars or elects to disable the overlay themselves?

schema/url/standalone-map/plugin-floating-boxplot.raml
```diff
@@ -0,0 +1,39 @@
...
+
+types:
+  # Type name determines the name of the Java class that will be generated.
+
+  FloatingBoxplotPostRequest:
+    type: DataPluginRequestBase
+    properties:
+      config: FloatingBoxplotSpec
+
+  # think this always gives outliers and means, saves some user configuration?
```
asizemore, line 11: this is a great point. Do we need the full power of each viz in the map? To keep things simple, can we lean toward as little user configuration as possible? A general question

schema/url/standalone-map/plugin-floating-boxplot.raml
```diff
@@ -0,0 +1,39 @@
...
+    properties:
+      outputEntityId:
+        type: string
+      xAxisVariable: 
+        type: VariableSpec
+      yAxisVariable: 
+        type: VariableSpec
+      overlayConfig?:
+        type: OverlayConfig
+      maxAllowedDataPoints:
```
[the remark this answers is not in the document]
d-callan: if the request would return greater than that number of points, we say sorry instead.

schema/url/standalone-map/plugin-floating-lineplot.raml
asizemore: for the sam, could we break up line into two in order to simplify plot config? One for categorical and one for continuous?

schema/url/standalone-map/plugin-floating-boxplot.raml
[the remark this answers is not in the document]
asizemore: oh just saw the other comment about this. Nvm :)

dmgaldi (approved):
I didn't look too closely at all of the floating plugins, but the spec looks good to me!

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/BoxplotPlugin.java
```diff
@@ -127,7 +127,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
           "plot.data::box(" + DEFAULT_SINGLE_STREAM_NAME + ", variables, '" +
               showPoints + "', " +
               showMean + ", " +
-              computeStats + ", '" +
+              computeStats + ", NULL, TRUE, TRUE, '" +
```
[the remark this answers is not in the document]
asizemore: no wait should be 
"', NULL, TRUE, TRUE, '"

adding the first apostrophe

src/main/java/org/veupathdb/service/eda/ds/plugin/pass/LineplotPlugin.java
```diff
@@ -145,7 +145,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
                                         ", viewport" + 
                                         ", " + numeratorValues +
                                         ", " + denominatorValues + 
-                                        ", '" + deprecatedShowMissingness + "')";                          
+                                        ", NULL, TRUE, TRUE, '" + deprecatedShowMissingness + "')";
```
asizemore, line 148: also needs another single apostrophe i think

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/FloatingLineplotPlugin.java
```diff
@@ -0,0 +1,160 @@
...
+    return List.of(VECTORBASE_PROJECT);
+  }
+
+  @Override
+  public ConstraintSpec getConstraintSpec() {
+    return new ConstraintSpec()
+      .dependencyOrder(List.of("yAxisVariable"), List.of("xAxisVariable"), List.of("overlayVariable"))
+      .pattern()
+        .element("yAxisVariable")
+          .description("Variable must be a number and be of the same or a child entity as the X-axis variable.")
```
asizemore, line 53: Description out of date - we dont require the y axis var to be a number

schema/url/standalone-map/plugin-floating-lineplot.raml
bobular: That would be an option - let's see how it goes with what we have.  I think I am about to do a major lineplot refactor... :grimacing:

PR #249, opened 2023-03-12
Simplify class declarations and first shot at scatterplot superclass

ryanrdoherty:
@d-callan , don't be fooled by the name of the branch.  These changed stemmed from that conversation you and Dave and I had about trying to merge the scatterplot plugins into one.  I discovered it's not that easy, but maybe this is a good first step.  Then while I was in there, I made the change from 2 or 3 required class declaration methods down to 1, so hopefully that can reduce code.  Sorry- I know it's not much to make the plugin writing easier.

Basically the questions are:
1. Do you want me to roll back the scatterplot factor or do you think it's nice and can be grown?
2. Is the constraint spec ok to apply to all three scatterplots (the ones with computes)?


d-callan:
i really appreciate you giving this some time. ive been thinking a lot about what a refactor might look like as well, and realized that if we take the time to refactor we should probably do it in a way that gets us ready for things like other types of maps or network diagrams as well. at the very least, i think we need to be mindful those things are coming and consider how/ where they fit into this world before investing too much energy into a refactor. 

another thing to consider is that the SAM we are now anticipating writing different descriptions for what are essentially the same constraints, bc the ux/ context for choosing variables is different there. (EDA weve assumed a user interacts w axes selections first, SAM theyll mostly likely have an overlay chosen first.)

id be happy to leave this a draft for now, maybe to serve as a place to collect some thoughts/ opinions. for ex, id love to hear more about what made this more complicated than originally anticipated.

ryanrdoherty:
Decided to ditch the scatterplot superclass for now- we can easily try again later when requirements are more clear.  For now, just simplifying the class declaration methods for code reduction and removing an unused interface method.

PR #251, opened 2023-03-21
Some raml cleanup and dealing with date variables

dmgaldi:


schema/url/standalone-map/plugin-standalone-map-markers.raml
```diff
@@ -34,13 +32,64 @@ types:
...
+      binEnd?: string
+
   BinRange:
     additionalProperties: false
     properties:
-      binStart?: string # number | date
-      binEnd?: string # number | date
       binLabel: string
 
+  OverlayType:
```
d-callan, line 56: i think my only question/ comment so far is about validation.. it seems like this puts more pressure on us to make sure people cant do things like pass us a date variable spec as overlay and then claim the overlay type is numeric or something?

schema/url/standalone-map/plugin-standalone-map-markers.raml
```diff
@@ -34,13 +32,64 @@ types:
       mapElements: StandaloneMapElementInfo[]
 
   # TODO remove once quantile endpoint is merged.
+  NumericBinRange:
```
d-callan, line 35: @ryanrdoherty and I were just talking about maybe using HistogramBin instead of BinRange in #240 and i wonder how this change impacts that

ryanrdoherty: Not sure.  Might be handy to use the same bin type as the distribution endpoint since then we can write methods to handle both.  Am I wrong that the efficiency of number vs string even when we know it's a number is negligible since this is just the bin definitions and not something for each row in a result?  The plugin can always coerce into the right type for processing and gain the efficiency benefits there.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/OverlaySpecification.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.veupathdb.service.eda.generated.model.OverlayType.CATEGORICAL;
+
+public class OverlaySpecification {
+  private final List<String> labels;
+  private final List<NormalizedBinRange> binRanges;
+  private OverlayRecoder overlayRecoder;
+
+  public OverlaySpecification(OverlayConfig overlayConfig, Function<VariableSpec, String> varTypeFinder) {
+    final String variableType = varTypeFinder.apply(overlayConfig.getOverlayVariable());
```
d-callan, line 22: is there anything stopping someone from trying something like claiming a continuous overlay type and passing bin start and ends and labels like 'blue' ? i mean i dont think well actually see that, but .. idk thought it might be worth mentioning

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/FloatingScatterplotPlugin.java
```diff
@@ -101,7 +102,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
     varMap.put("overlay", spec.getOverlayVariable());
     String valueSpec = spec.getValueSpec().getValue();
     String yVarType = util.getVariableType(spec.getYAxisVariable());
-    String overlayValues = listToRVector(spec.getOverlayValues());
+    String overlayValues = listToRVector(spec.getOverlayValues().stream().map(BinRange::getBinLabel).collect(Collectors.toList()));
```
d-callan, line 105: so right now the R supporting this doesnt know about continuous overlays, but that is going to have to change very soon. something for me to think about mostly i guess...

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/StandaloneMapMarkersPlugin.java
```diff
@@ -112,94 +104,55 @@ void addRow(double lat, double lon, String overlayValue) {
...
+      case CONTINOUS -> validateContinousBinRanges((ContinousOverlayConfig) overlayConfig);
+      case CATEGORICAL -> validateCategoricalOverlayValues((CategoricalOverlayConfig) overlayConfig);
     }
+  }
 
-    // start is always before end
-    boolean anyStartAfterEnd = binRanges.stream().filter(bin -> bin.getBinEnd() <= bin.getBinStart()).findFirst().isPresent();
-    if (anyHaveBinStart & anyStartAfterEnd) {
-      throw new ValidationException("All BinRanges must have binStart less than binEnd.");
+  private void validateCategoricalOverlayValues(CategoricalOverlayConfig overlayConfig) throws ValidationException {
```
d-callan, line 135: this answers a question i had somewhere else then.. except it might be okay to pass some cont vars as a categorical type overlay. itd be odd i guess, but not invalid, to say choose your 7 favorite values from some int w 12 unique values.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/StandaloneMapMarkersPlugin.java
```diff
@@ -112,94 +104,55 @@ void addRow(double lat, double lon, String overlayValue) {
...
-          prevBinEnd = binEdges.get(binStart);
-          first = false;
-        } else {
-          if (prevBinEnd > binStart) {
-            throw new ValidationException("Bin Ranges must not overlap");
-          }
-          prevBinEnd = binEdges.get(binStart);
-        }
-      }
+    int numDistinctOverlayVals = new HashSet<>(overlayConfig.getOverlayValues()).size();
```
d-callan, line 140: i feel like some of this validation isnt unique to map markers, but just the overlay generally, which will be used for all floating charts as well.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/StandaloneMapMarkersPlugin.java
```diff
@@ -112,94 +104,55 @@ void addRow(double lat, double lon, String overlayValue) {
...
-      }
+  private void validateContinousBinRanges(ContinousOverlayConfig overlayConfig) throws ValidationException {
+    if (!getUtil().getVariableDataShape(overlayConfig.getOverlayVariable()).equalsIgnoreCase(APIVariableDataShape.CONTINUOUS.toString())) {
+      throw new ValidationException("Input overlay variable %s is %s, but provided overlay configuration is for a conginuous variable"
+          .formatted(overlayConfig.getOverlayVariable().getVariableId(), getUtil().getVariableDataShape(overlayConfig.getOverlayVariable())));
+    }
+    boolean anyMissingBinStart = overlayConfig.getOverlayValues().stream().anyMatch(bin -> bin.getStart() == null);
+    boolean anyMissingBinEnd = overlayConfig.getOverlayValues().stream().anyMatch(bin -> bin.getEnd() == null);
+    if (anyMissingBinStart || anyMissingBinEnd) {
+      throw new ValidationException("All numeric bin ranges must have start and end.");
```
d-callan, line 154: overlapping ranges?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/OverlaySpecification.java
dmgaldi: Would it be better if we just inferred the labels as "${start}-${end}" for continuous? I'm not sure whether the client will need to specify labels for bin ranges

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/FloatingScatterplotPlugin.java
dmgaldi: Ah yeah, I just made the minimal possible changes to the floating plugins to get everything compiling for now

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/StandaloneMapMarkersPlugin.java
dmgaldi: Ah I see, I'm a little behind on the requirements for the floating plugins. Are there existing mock-ups that you know of?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/OverlaySpecification.java
d-callan: Good question. Right now the R fxn that makes the bins will provide exact values for start and end, but more friendly/ rounded values in the labels.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/StandaloneMapMarkersPlugin.java
d-callan: Hahahahaha... No.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/StandaloneMapMarkersPlugin.java
d-callan: Any bins I make will be.. and it doesn't seem crazy to ask custom bins to be. I thought I had added something for overlapping ranges already tbh

PR #255, opened 2023-04-04
add geohash and count to response

d-callan:


bobular (approved):
Not sure I'm going to get this tested. Running into gradle errors now.

It's the fix I would have made though!

bobular:
Managed to test it too. It works!

PR #258, opened 2023-04-12
organize standalone map floating charts into groups/ apps

d-callan:


d-callan:
@ryanrdoherty i wonder how much itd cost to have an arbitrary numbr of 'tiers' to the organization of these vizs... like `standalone-map/distributions` instead

adnauseum (approved):
LGTM! 🚀

ryanrdoherty:
> @ryanrdoherty i wonder how much itd cost to have an arbitrary numbr of 'tiers' to the organization of these vizs... like `standalone-map/distributions` instead

I don't quite understand what the goal is of separating/organizing these into separate apps given they all apply to VB and all don't have a compute.  As we've discussed, these are the only reasons the server cares about apps as a concept, and we were thinking about:
1. Moving the applicable projectId configuration into the client
2. Making a visualization accept a class of computes (all returning the same data API) and taking the compute name as part of the compute config (rather than having the app supply/dictate it).

So in that sense, they could both be flat or be whatever organization you want.  I feel like the layers of url segments affects the client more since they will have to know the full URL of each viz (with some arbitrary number of segments).  But maybe that is fine too.

ryanrdoherty (approved):

d-callan:
currently we organize them into groups w different descriptions and display names, which is exposed via the apps/ metadata endpoint. so the only real reason to do it for the map is to be consistent w what we did in other places (mbio being the primary ex i guess). the other obvious option is to decide we dont want to be consistent, and just let the client decide for itself how it wants to organize them.

bobular:
I'm fine with organising by type.  I'm assuming the client already knows how to do this for mbio so it may just require an environment variable (SINGLE_APP_MODE IIRC) unsetting?

Perhaps we can nail down a client development plan before merging this?

The client code does expect just a single level of hierarchy so maybe best to keep it that way.
```
        path: `/apps/${computationName}/visualizations/${visualizationName}`,
```

bobular:
Hi @adnauseum 

I've looked at the client-side some more.

Currently `MapAnalysis.tsx` and `MapVizManagement.tsx` are using a single `app` (hardcoded to be the 'pass' app).

When* we switch to hierarchical/organised apps this will need to be handled more like the EDA in `ComputationRoute.tsx` - so we'd need to obtain a list of apps that are allowed for `projectId` (as already done in `ComputationRoute.tsx`) and pass that to `MapVizManagement`.  I think we can re-use `StartPage.tsx` (possibly adapted to work in a narrower container) to do the layout, instead of the hardcoded grouping in `NewVisualizationPickerGrouped`.

(* I think "when" rather than "if". Long term we'll want the option to have computation apps in the map I think. Perhaps for abundance calculations, or maybe for species distribution modelling/interpolation --> raster overlays)


@d-callan - I think this will be easiest to rewire while the back end is still a branch. I am happy to do it after finishing [wiring up the new markers endpoint](https://github.com/VEuPathDB/web-monorepo/pull/45). Does that impact on anyone else?

dmfalke:
@bobular and I were chatting about this the other day. Using `VectorBase` as the "project id" means that we can't use the grouping in Clinepi. Maybe that's okay, but we could also use a different, non-website specific, project id, such as `MapApp`.

Thoughts? @ryanrdoherty @d-callan

d-callan:
If project Id isn't a list I feel it should be

bobular (approved):

ryanrdoherty:
All this reminds me of our conversation about moving this hierarchy and project IDs out to the client, with the philosophy that anything the service doesn't need to know to do what it does is not necessary unless only the service can know how to provide that info (e.g. if it's read from a DB).  I sorta stand by the idea that an "app" (from the service's standpoint) is just a collection of visualizations that depend on a particular compute.  And that is with an eye toward a future where they depend on a compute result API (which multiple plugins could implement).  From that perspective data service plugins could be a flat list, though I do think it makes sense to group them essentially by compute until we figure out how do this "by compute result API" business.  Any other organization could be determined and owned by the client.

bobular:
Hi Ryan - will need to re-read and think about what you wrote, but for the record I added a client ticket to make sure we don't forget to deal with these issues

https://github.com/VEuPathDB/web-monorepo/issues/161

d-callan:
@ryanrdoherty fwiw I haven't been a fan of most of the work I've done in the data service for the SAM. But I also think reimagining how this works isn't something we can meaningfully do right now. I was also daydreaming about a day where we abstract away from the concept of visualizations entirely on the backend and make this a stats centric service. So it warrants some discussion I think around what we'd like to start doing when we stop doing what we're currently doing, but for now...

dmfalke:
I agree with Ryan's idea. That's basically how the client side of this is designed. The same viz react component is used with multiple computes/apps

d-callan:
So the backend has abstract plot classes, which are reused across application and context specific viz plugins. That's the current design. I agree, in general, that there is no real reason the backend needs to know the context/ things like the display name or organization of vizs etc. But I don't know if Ryan's idea goes far enough tbh. I've spent three or so years building these plugins and I am not convinced the backend actually needs to know about vizs at all. Why shouldn't I just return raw stats?

dmfalke:
@d-callan I didn't mean to dismiss your thoughts. I don't really understand what you are proposing. Can you give an example, e.g., what would the client ask for if it wants a histogram?

d-callan:
I haven't thought about it enough to have an exact proposal, but roughly I think what I'd like is to expose the underlying functions in plot.data more or less directly. There are stats there that can be done per plot panel, per group/color per panel or per bin per group per panel. The histogram has a related function for counts per bin, which we could just expose. Then how the client wants to build vizs becomes much more flexible.

PR #267, opened 2023-05-02
Add differential abundance volcanoplot

asizemore:
Adds a plugin for the differential abundance volcano plot.

This one is funny because we don't actually request any variables for the volcano. Instead, all of the data for the plot is supplied by the compute stats result. No overlays, no facets, no missingness. So i think we can remove the `validateVisualizationSpec` and `getConstraintSpec` methods? Also since we have the `getComputeResultsStats` do we even need `getRequestedStreams`?


schema/url/differentialabundance/plugin-differentialabundance-volcanoplot.raml
```diff
@@ -0,0 +1,17 @@
...
+    type: DataPluginRequestBase
+    properties:
+      computeConfig: DifferentialAbundanceComputeConfig
+      config: DifferentialAbundanceVolcanoplotSpec
+
+  # No variables, no missingness
+  DifferentialAbundanceVolcanoplotSpec:
+    additionalProperties: false
+    properties:
+      outputEntityId:
```
d-callan, line 16: lol. they cant exactly change the output entity, since this plot doesnt belong in an entity space that actually exists in the entity diagram. is it possible this has no viz config?

src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java
```diff
@@ -83,6 +84,10 @@ public class AppsMetadata {
           "Visualize between-sample (beta) comparisons in microbial diversity, using Bray-Curtis dissimilarity, Jensen-Shannon Divergence, or the Jaccard Index",
           List.of(MICROBIOME_PROJECT),
           viz("scatterplot", new BetaDivScatterplotPlugin())),
+      app("differentialabundance", "Differential Abundance", "differentialabundance",
+          "Find taxa or genes that are differnetially abundant between two groups.",
```
d-callan, line 88: typo

src/main/java/org/veupathdb/service/eda/ds/plugin/differentialabundance/DifferentialAbundanceVolcanoplotPlugin.java
```diff
@@ -0,0 +1,91 @@
...
+  public List<String> getProjects() {
+    return List.of(AppsMetadata.MICROBIOME_PROJECT);
+  }
+
+  @Override
+  protected ClassGroup getTypeParameterClasses() {
+    return new ClassGroup(DifferentialAbundanceVolcanoplotPostRequest.class, DifferentialAbundanceVolcanoplotSpec.class, DifferentialAbundanceComputeConfig.class);
+  }
+
+  // // I think we can remove this whole thing ?!? there are no variables for the user to select
```
d-callan, line 44: we need @ryanrdoherty .. bc i think outside of giving this viz a name and description this is just a pass-through endpoint for the stats endpoint from compute.

src/main/java/org/veupathdb/service/eda/ds/plugin/differentialabundance/DifferentialAbundanceVolcanoplotPlugin.java
```diff
@@ -0,0 +1,91 @@
...
+        .setIncludeComputedVars(true)
+    );
+  }
+
+  @Override
+  protected void writeResults(OutputStream out, Map<String, InputStream> dataStreams) throws IOException {
+    DifferentialAbundanceVolcanoplotSpec spec = getPluginSpec();
+    DifferentialAbundanceStats stats = getComputeResultStats(DifferentialAbundanceStats.class);
+
+    System.out.println(stats);
```
ryanrdoherty, line 84: You need to convert the stats object to a json string and write it to the output stream.  This is not too hard, but maybe I'll add a utility if we think a direct stream from compute stats out the door will be a common use case?

ryanrdoherty (changes requested):

src/main/java/org/veupathdb/service/eda/ds/plugin/differentialabundance/DifferentialAbundanceVolcanoplotPlugin.java
d-callan: i think its likely to be a common case. i need to think some more about it to be sure, but i could easily see us doing the same for the correlations app(s)

src/main/java/org/veupathdb/service/eda/ds/plugin/differentialabundance/DifferentialAbundanceVolcanoplotPlugin.java
ryanrdoherty: I wrote up a utility (above) and used it (below) to do this and simplified the other required methods.  If we really think this is a common case maybe we can make a superclass for all the "no-config, stats-stream" viz plugins.  Future TBD.

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
```diff
@@ -141,8 +141,9 @@ public final AbstractPlugin<T,S,R> processRequest(String appName, T request, Ent
       throw new BadRequestException("Compute results are not available for the requested job.");
     }
 
-    // if plugin requires a compute, get computed var metadata
-    List<VariableMapping> computedVars = _computeInfo.isPresent()
+    // if plugin requires a compute and metadata is expected for this compute, get computed var metadata
+    boolean expectMetadata = !appName.equals("differentialabundance");  // Diff abundance has no computed var metadata
```
ryanrdoherty, line 145: I think there's a better way to do this- we could either have the plugin declare whether they expect to have var metadata (always false for non-compute plugins) or we can handle the 404 and return emptyList() in that case.  The first way forces the plugins to override a method if they have computes that don't provide vars; the second way means we're making that service call unnecessarily just to get a 404 back.  So I think the first way is probably better.

ryanrdoherty: Also if the data plugin thinks there should be vars but we get a 404 that should throw an error to alert someone to the mismatch.

ryanrdoherty (approved):

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
ryanrdoherty: Added here: https://github.com/VEuPathDB/EdaDataService/pull/267/commits/d8862e724789689d8c59858a4d4d66c1d102e1ff

asizemore: Thanks @ryanrdoherty ! That makes a lot of sense

api.raml
```diff
@@ -371,6 +371,21 @@ securitySchemes:
               application/json:
                 type: lib.ScatterplotPostResponse
 
+  /differentialabundance/visualizations:
+    displayName: Visualizations for interrogating differentially abundant taxa, genes, or pathways
```
d-callan, line 375: Should just say taxa for now?

d-callan (approved):

api.raml
asizemore: The EC numbers are enzymes so i was hoping to dump them into the "genes" umbrella

PR #269, opened 2023-05-03
Include count in the marker regardless of value spec

### LINKED ISSUE #261: New map-markers service should return counts alongside proportions
When in `valueSpec = 'proportion'` mode, please could the response's `overlayValues` also include the `n` entity counts alongside the existing `value`?  The SAM client needs this to calculate the number of entities in the viewport.

Proportion mode is currently disabled because we don't have the numbers we need.

Related to https://github.com/VEuPathDB/web-eda/issues/1674 (where we decided to leave this to the new endpoint, but I don't think we made a ticket)

dmgaldi:
Resolves #261

This changes is a little awkward from the model perspective, but this is a backwards-compatible implementation that includes count when proportion is requested.

Tested against mega study, requested both value specs.

### DIFF
--- a/schema/library.raml
+++ b/schema/library.raml
@@ -1792,13 +1792,18 @@ types:
     properties:
       overlayValues:
         type: string[]
+  RangeWithCountAndValue:
+    type: BinRangeWithValue
+    additionalProperties: false
+    properties:
+      count: number
   StandaloneMapElementInfo:
     type: object
     additionalProperties: false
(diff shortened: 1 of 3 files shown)


ryanrdoherty (approved):
Looks good.  I'll apply the "common" types in the derived vars branch once this is merged to master.

d-callan (approved):

PR #270, opened 2023-05-04
force dates to be dates when they are dates

asizemore:
Resolves part of https://github.com/VEuPathDB/veupathUtils/issues/14

This does not fix all the issues with the offending variables, but it at least ensures the vars will have the appropriate type when they get sent into `getDiscretizedBins`.


asizemore:
note - sometimes they are date-times. This would force them all to dates and drop the time. Are we okay with that?

src/main/java/org/veupathdb/service/eda/ds/plugin/filteredmetadata/ContinuousVariablePlugin.java
```diff
@@ -88,13 +88,19 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
...
+          List<String> requestedMetadata = spec.getMetadata();
 
           useRConnectionWithRemoteFiles(Resources.RSERVE_URL, dataStreams, connection -> {
             connection.voidEval(util.getVoidEvalFreadCommand(DEFAULT_SINGLE_STREAM_NAME, spec.getVariable()));
             // for convenience
             connection.voidEval("x <- " + DEFAULT_SINGLE_STREAM_NAME + "$" + util.toColNameOrEmpty(spec.getVariable()));
 
+            // Dates are read in as class=character. If we see the variable type is date, let's
+            // force the data in R to be have class=date. See veupathUtils #14 for more details.
+            if (util.getVariableType(spec.getVariable()) == "DATE") {
```
d-callan, line 100: There's probably a better way to check this using like APIVariableType or something

d-callan (approved):

asizemore:
This var comes in with a time. It's always the same midnight time so it's not containing any useful information for this var.

d-callan:
I say ignore the time for now, well handle that separately as needed. Thanks

Issue #275, opened 2023-05-11
Visualizations: Relax dependency order constraint


d-callan:
Right now we define a dependency order of visual elements within any given viz. this looks something like [yaxis], [xaxis], [overlay, facet] typically. what this does is require the xaxis variable to be the same or a parent entity of the yaxis variable, and the overlay and facet variables to belong to the same or parent entity of the xaxis variable in turn. this is done to prevent duplicated values from appearing in the summary statistics of any visualization we produce.

Its possible we can revise these to be something like [yaxis], [xaxis, overlay, facet] ?? 

This would be useful for all vizs we make w eda, but important in particular for the map since we have some pretty critical terms (like collection start date) on very ancestral entities which we likely want to plot w the assay data (a leaf entity).

PR #276, opened 2023-05-12
updated data element dependency order

d-callan:
see https://github.com/VEuPathDB/EdaDataService/issues/275 for details


bobular (approved):
I approve in concept. Are we sure every single back end request will not 500? If so, why didn't we do this months ago?

d-callan:
I don't think they'll 500, but I'm also not confident I have the logic right to ensure we aren't finding summary stats over dups. Plus now the dep order is different in different plugins. It's not consistent across different scatter plots I think for ex, based on context. So I hadn't done it yet bc I was less sure how to do it well or if the potential inconsistency was worth it. But the map seems to be making the argument in favor.

asizemore (approved):

PR #289, opened 2023-06-20
Floating timeseries

### LINKED ISSUE #263: Map: Timeline plugin
Once a long long time ago, a silly Danielle implemented a time series plugin we never used for the pass app. We need to do the same now for the map floaters.

d-callan:
resolves #263


src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/FloatingTimeSeriesPlugin.java
```diff
@@ -0,0 +1,39 @@
...
+  @Override
+  public ConstraintSpec getConstraintSpec() {
+    return new ConstraintSpec()
+      .dependencyOrder(List.of("yAxisVariable"), List.of("xAxisVariable", "overlayVariable"))
+      .pattern()
+        .element("yAxisVariable")
+          .description("Variable must be of the same or a child entity as the X-axis variable.")
+        .element("xAxisVariable")
+          .temporal(true)
+          .description("Variable must be temporal and be the same or a child entity as the variable the map markers are configured with, if any.")
```
bobular, line 34: I know this is the general wording for previous plugins but I'm still finding it a bit tricky to understand.

"Variable must be temporal and belong to the same or child entity as the variable the map markers are configured with, if any"

bobular (approved):
Looks good! One verbiage suggestion you can take or leave.

PR #291, opened 2023-06-27
Standalone map marker implementation

dmgaldi:
## Overview
* Initial attempt at bubble plugin.
* Currently, no way for client to interpret relative size/gradient of color and entity count fields
* I'm not sure if color is the right name for the field or if that's more of a UI detail?

## Testing
* Tested new and old markers with the mega study


src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/CategoricalRatioAggregator.java
```diff
@@ -0,0 +1,40 @@
+package org.veupathdb.service.eda.ds.plugin.standalonemap.markers;
+
+import java.util.Set;
+
+/**
+ * Map marker aggregator taking categorical variable and providing a ratio of number of elements with values in a
+ * set of numerator values over number of elements with values in a set of denominator values.
+ */
+public class CategoricalRatioAggregator implements MarkerAggregator<Double> {
```
d-callan, line 9: frustratingly, id like to name this something other than ratio.. ideally proportions maybe. but we also have proportions in another context w a slightly different semantic intent..

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/CategoricalRatioAggregator.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public void addValue(String s) {
+    if (s == null) {
+      return;
+    }
+    if (numeratorValues.contains(s)) {
+      numNumeratorMatches++;
+    }
+    if (denominatorValues.contains(s)) {
```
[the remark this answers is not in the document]
d-callan: sry if i didnt say this before...

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/CategoricalRatioAggregator.java
d-callan: maybe we can figure how to collapse these two categorical aggregators into one?

d-callan: or reorganize them somehow? im not sure what i want yet, sorry..

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/ContinuousAggregators.java
```diff
@@ -0,0 +1,130 @@
...
+      sum += Double.parseDouble(d);
+      n += 1;
+    }
+
+    @Override
+    public Double finish() {
+      if (n == 0) {
+        return null;
+      }
+      return sum / n;
```
d-callan, line 30: how many digits will this report to? should we format or the client? i dont remember what we do other places..

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/MapBubbleSpecification.java
```diff
@@ -0,0 +1,49 @@
...
+  private final VariableSpec overlayVariable;
+  private OverlayRecoder overlayRecoder;
+  private Supplier<MarkerAggregator<Double>> aggregatorSupplier;
+
+  public MapBubbleSpecification(ColorConfig overlayConfig,
+                                Function<VariableSpec, String> varShapeFinder) {
+    this.overlayVariable = overlayConfig.getOverlayVariable();
+    if (CATEGORICAL.equals(overlayConfig.getOverlayType())) {
+
+      if (!varShapeFinder.apply(overlayVariable).equalsIgnoreCase(APIVariableDataShape.CATEGORICAL.getValue())) {
```
d-callan, line 25: this should probably allow ordinals as well. theyre a valid subcategory of categorical.

d-callan (commented):
i think the only thing id say really should change before merge is to enforce our categorical ratios for bubbles are properly proportions, by requiring all numerator values to be in the denominator as well.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/CategoricalRatioAggregator.java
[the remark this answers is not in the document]
dmgaldi: No worries, that makes sense. I couldn't imagine meaningful results where they weren't proportions.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/CategoricalRatioAggregator.java
dmgaldi: I did some re-naming. Let me know if you have thoughts about where we ended up. I made them separate to try to reduce duplication with the different map marker plugins.

dmgaldi:
> i think the only thing id say really should change before merge is to enforce our categorical ratios for bubbles are properly proportions, by requiring all numerator values to be in the denominator as well.

Sounds good, I have this addressed in the latest revision.

PR #295, opened 2023-07-11
API and docs for collection marker

dmgaldi:
Adding a new API for collection map markers. The implementation is stubbed out for now. Hoping to just get feedback on the API first in order to make sure my understanding is solid.


schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,81 @@
...
+    additionalProperties: false
+    description: Specification for a map marker
+    properties:
+      outputEntityId:
+        required: true
+        type: string
+        description: Entity ID associated with collection variable of interest.
+      collectionVariableId:
+        required: true
+        type: VariableSpec
```
d-callan, line 20: we should have a CollectionSpec, iirc

schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,81 @@
...
+        description: Collection variable to use for markers.
+      memberVariables:
```
d-callan, line 22: id maybe just call it desiredMemberVariables or something, for clarity.

schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,81 @@
...
+        type: MemberVariable[]
+        description: | 
+          Optionally, a set of member variable IDs can be provided. Aggregate output values will be provided per marker
+          for each variable in this set. By default, all variables in collection will be provided. The maximum number of
+          variables we can provide output for per marker is 8.
```
d-callan, line 27: i wouldnt mind just requiring they pass a list of 8 or fewer and write some validation, rather than provide values for everything by default.

schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,81 @@
...
+
+  # Note that this is not using inheritance as it breaks API doc generation. Instead, both continuous/categorical
+  # configurations are top-level fields and the backend will validate that the config is compatible with the associated var.
+  MemberVariable:
+    additionalProperties: false
+    properties:
+      variableId: string
+      categoricalConfig:
```
d-callan, line 35: all members should share the same configuration here, so im not sure it needs to belong to MemberVariable. where you have MemberVariable[] above i think could just be VariableSpec[], then add this config to StandaloneCollectionMapMarkerSpec

schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,81 @@
...
+      entityCount: number
+      avgLat: number
+      avgLon: number
+      minLat: number
+      minLon: number
+      maxLat: number
+      maxLon: number
+      values: CollectionMemberAggregate[]
+
+  CollectionMemberAggregate:
```
d-callan, line 70: these should also provide confidence intervals and denominators. so maybe add lowerBound, upperBound and n

d-callan: we should talk about the confidence intervals probably, and how to compute them...

schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,81 @@
...
+  CategoricalMemberConfig:
+    additionalProperties: false
+    description: | 
+      Configuration for aggregating categorical values into a proportion. Note that numerator values must be a subset
+      of denominator values.
+    properties:
+      numeratorValues: string[]
+      denominatorValues: string[]
+
+  MapMarkerElement:
```
d-callan, line 56: should this be called CollectionMapMarkerElement?

schema/url/standalone-map/plugin-collection-markers.raml
dmgaldi: Oh I see, is this even the case for categorical variables? I'm not that familiar with collections, do all member variables of categorical collections share variable values?

d-callan: Yes it's true of categorical collections as well. It's perhaps the chief defining characteristic of collections that their members share theoretical values.

schema/url/standalone-map/plugin-collection-markers.raml
dmgaldi: Sounds good , maybe we should sync up on this tomorrow. Do we need an input confidence level value or do we use a reasonable default?

d-callan: We're using the default of 95% confidence everywhere else and should be here as well. We could add an input to keep us future proof if you'd like though.

schema/url/standalone-map/plugin-collection-markers.raml
```diff
@@ -0,0 +1,68 @@
...
+      maxLat: number
+      maxLon: number
+      values: CollectionMemberAggregate[]
+
+  CollectionMemberAggregate:
+    properties:
+      variableId: string
+      mean: number
+      median: number
+      confidenceInterval: NumberRange
```
d-callan, line 62: This is great. But I think we should also include the count of rows contributing to each bars/ variables summary value and confidence interval. Could call it `n` maybe, don't remember what we called it for lineplot..

d-callan (approved):

PR #296, opened 2023-07-17
collection markers

dmgaldi:
## Overview
Collection markers and some refactoring to re-use the quantitative overlay code from bubble markers. Testing is still pending data development work.


d-callan (commented):
have we worked out confidence intervals for medians and proportions yet? or need input? or maybe i just missed them?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/CollectionMapMarkersPlugin.java
```diff
@@ -45,8 +76,11 @@ protected AbstractPlugin<StandaloneCollectionMapMarkerPostRequest, StandaloneCol
 
   @Override
   protected void validateVisualizationSpec(StandaloneCollectionMapMarkerSpec pluginSpec) throws ValidationException {
-    final List<VariableDef> allMembers = getUtil().getCollectionMembers(pluginSpec.getCollection());
-    final Set<String> selectedMemberIds = pluginSpec.getSelectedMemberVariables().stream()
+    if (pluginSpec.getCollection() == null) {
+      throw new ValidationException("Collection information must be specified.");
+    }
+    final List<VariableDef> allMembers = getUtil().getCollectionMembers(pluginSpec.getCollection().getCollection());
+    final Set<String> selectedMemberIds = pluginSpec.getCollection().getSelectedMembers().stream()
```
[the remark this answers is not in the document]
d-callan: lol i see it.

schema/url/visualizations.raml
```diff
@@ -154,23 +154,26 @@ types:
...
     discriminator: overlayType
     properties:
       overlayType:
         type: OverlayType
-      overlayVariable:
-        type: VariableSpec
 
-  ContinuousQuantitativeOverlayConfig:
-    type: QuantitativeOverlayConfig
+  ContinuousAggregationConfig:
```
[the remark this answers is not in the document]
dmgaldi: This is used in the internal `QuantitativeAggregateConfiguration` class to determine what `MarkerAggregator` to user for a given request

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/MedianWithConfidenceAggregator.java
```diff
@@ -0,0 +1,59 @@
...
+    double upperIndex = values.size() * 0.5 + 1.96 * Math.sqrt(values.size() * 0.5 * 0.5);
+    double lowerBound;
+    double upperBound;
+    if (lowerIndex < 0) {
+      lowerBound = values.get(0);
+    } else {
+      lowerBound = values.get((int) Math.ceil(lowerIndex));
+    }
+    if (upperIndex >= values.size()) {
+      upperBound = values.get(values.size() - 1);
```
d-callan, line 45: i think what we are doing in R when the lowerindex < 0 or upperindex > size is to not return a confidence interval and produce a warning that we couldnt reliably calculate one or something. seems to me that really theres no conceptual reason why the confidence interval has to fall within the observed values, particularly when n is small.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/ProportionWithConfidenceAggregator.java
```diff
@@ -0,0 +1,56 @@
...
+      denominatorMatches++;
+    }
+  }
+
+  @Override
+  public AveragesWithConfidence finish() {
+    final double proportion = (double) numeratorMatches / denominatorMatches;
+    final double confidence = StatUtils.Z_SCORE_95 * (Math.sqrt((proportion * (1 - proportion)) / n));
+    final double upperBound = proportion + confidence;
+    final double lowerBound = proportion - confidence;
```
d-callan, line 44: oddly enough, here i think we can cap lower and upper bounds at 0 and 1 since we know all proportions have to fall within that range.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/MedianWithConfidenceAggregator.java
dmgaldi: Ah makes sense. I'll just return a null confidence interval. I wonder if we should add something to the model to indicate warnings?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/ProportionWithConfidenceAggregator.java
dmgaldi: Got it, I'll fix this up

Issue #294, opened 2023-07-07
Bubble markers validate overlay configuration


dmgaldi:
## Overview
The following request gives a 5xx instead of 4xx. It should give a 4xx because the categorical overlay config is missing numerator and denominator values.

```
{
  "studyId": "SCORESL01-1",
  "filters": [],
  "config": {
    "geoAggregateVariable": {
      "entityId": "EUPATH_0035127",
      "variableId": "EUPATH_0043203"
    },
    "latitudeVariable": {
      "entityId": "EUPATH_0035127",
      "variableId": "OBI_0001620"
    },
    "longitudeVariable": {
      "entityId": "EUPATH_0035127",
      "variableId": "OBI_0001621"
    },
    "overlayConfig": {
      "overlayType": "categorical",
      "overlayVariable": {
        "variableId": "ENVO_00000009",
        "entityId": "EUPATH_0035127"
      }
    },
    "outputEntityId": "EUPATH_0035127",
    "valueSpec": "count",
    "viewport": {
      "latitude": {
        "xMin": -86.57422361983717,
        "xMax": 86.53194966939856
      },
      "longitude": {
        "left": -180,
        "right": 180
      }
    }
  }
}
```

PR #300, opened 2023-08-01
Fix validation for bubble markers plugin

dmgaldi:
Make sure categorical overlays work for all categorical variable types. Previously, binary variables were getting rejected by validation.

This resolves https://github.com/VEuPathDB/EdaDataService/issues/294


bobular (commented):
Just "reviewed" after the event.

Sorry I missed this 9 days ago!

On the plus side, I can say that it's now working on dev for binary variables that used to fail!

Looks nice!

PR #305, opened 2023-08-16
Entity count and marker data represents overlay value presence

dmgaldi:
Resolves https://github.com/VEuPathDB/EdaNewIssues/issues/673

Added a new method to aggregator interface to determine if the overlay it represents applies to a record.


src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/BubbleMapMarkersPlugin.java
```diff
@@ -146,7 +147,9 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
       mapEle.setMaxLat(data.getMaxLat());
       mapEle.setMinLon(data.getMinLon());
       mapEle.setMaxLon(data.getMaxLon());
-      mapEle.setOverlayValue(data.getMarkerAggregator().finish());
+      if (data.getMarkerAggregator() != null) {
+        mapEle.setOverlayValue(data.getMarkerAggregator().finish());
```
bobular, line 151: Hi @dmgaldi and @d-callan 
I'm still seeing missing `overlayValue` in the bubbles response, and the code here confirms this.
Don't we want this to be a little filter (if there is no overlay data to aggregate)?

Should the `output.add(mapEle);` be inside the `if (data.getMarkerAggregator() != null)` condition?

bobular: I see the missing `overlayValue` when the denominator values are hand-picked (instead of "all"). 

Following the original bug report, with SCORE S. mansoni, and "Hookwork by microscopy", choose "positive/positive")

I will look deeper at the code here

bobular: Or maybe it's the `appliesTo` which needs looking at?

https://github.com/VEuPathDB/EdaDataService/blob/704a80fce4cd600b07b7b08bc00e32d66a7dea90/src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/aggregator/CategoricalProportionAggregator.java#L39-L41

PR #309, opened 2023-08-18
Support dates for bubble markers

dmgaldi:


bobular:
I've started doing the front end work for this.

bobular:
I've put it as "awesome for beta" but can upgrade/downgrade as appropriate

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/aggregator/MedianAggregator.java
```diff
@@ -0,0 +1,97 @@
+package org.veupathdb.service.eda.ds.plugin.standalonemap.aggregator;
+
+import java.util.Comparator;
+import java.util.PriorityQueue;
+import java.util.function.Function;
+
+public class MedianAggregator implements MarkerAggregator {
```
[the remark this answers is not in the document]
ryanrdoherty: Also would help to have unit tests specifically for the median util.  This is complex enough to need thought so some proof of logic would be helpful.

src/main/java/org/veupathdb/service/eda/ds/utils/CommonFormats.java
```diff
@@ -0,0 +1,16 @@
+package org.veupathdb.service.eda.ds.utils;
+
+import java.time.LocalDate;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+
+public class CommonFormats {
+  public static final DateTimeFormatter TABULAR_DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
ryanrdoherty, line 8: I thought we used to output date cols as datatimes (yyyy-MM-ddThh:mm:ss).  Maybe we don't do this any more?

ryanrdoherty (approved):

src/main/java/org/veupathdb/service/eda/ds/utils/CommonFormats.java
bobular: Can of worms there!

The client hasn't been updated to handle bubble dates yet, so the best thing is for the back end to do the most consistent thing and we will take care of whichever format (we'll likely not display any time component to the user, but no problem handling the full date-time in the response and subsequent calculations)

bobular:
@dmfalke and I discussed how best to handle this on the client side.

We decided that the quite extensive client work (date and number handling is always a bit of a faff) should be done on Dave's refactored branch that will not be merged for b65.

However, we appreciate that we don't want this PR to get stale and we don't want 500 errors for the user.

So we've done some temporary client work to avoid 500 errors and temporarily disable date vars in the client (for bubble markers)

https://github.com/VEuPathDB/web-monorepo/pull/453

d-callan (approved):

bobular:
To test this against the client, I'll need this updating from `master` please.  I was getting 404s from merging-internal when running this branch locally alongside the other services on `latest`.

Issue #307, opened 2023-08-17
Map Impute Zeroes: update backend plugins to impute zeroes on tall megastudies


d-callan:
depends on https://github.com/VEuPathDB/EdaSubsettingService/issues/113

PR #311, opened 2023-08-21
study specific vocabs in SAM floaters

d-callan:
resolves https://github.com/VEuPathDB/EdaDataService/issues/307

mostly just putting together my thoughts so far


src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
```diff
@@ -610,4 +626,53 @@ public String getRBinListAsString(List<String> labels) {
...
+    return true;
+  }
+
+  public String getRStudyVocabsAsString(VocabByRootEntityPostResponse studyVocab) {
+    // TODO make a string to build the vocabs slot
+  }
+
+  public Strng getRStudyVocabsAsString(List<VocabByRootEntityPostResponse> studyVocabs) {
+    // TODO loop through and call the other version once for each entry. sandwich that in a R list.
+  }
```
dmgaldi, line 640: I'm rolling off some other work. If you want I can take a stab at filling in the AbstractPlugin methods. What's the desired string format here though? I'm having trouble conceptualizing it since the vocab is two dimensional

d-callan: this is still wip, and i dont think will be awful to do. i just need to get to it. if youre in need of work, mbio is looking for help still.

src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
```diff
@@ -610,4 +641,138 @@ public String getRBinListAsString(List<String> labels) {
...
+    // TODO check a bunch of stuff i no longer remember
+    return true;
+  }
+
+  public String getRStudyVocabsAsString(DynamicDataSpecImpl dataSpec, ResponseFuture studyVocab) {
+    PluginUtil util = getUtil();
+    
+    // this assuming the first ancestor is the root one is a bit hacky, but i did the same in R so...
+    // obviously doing the same awful thing twice makes it ok. but doing any better is higher cost than i have time for right now :(
+    String readStudyVocabInR = "data.table::fread(" + studyVocab + ", header = FALSE)";
```
dmgaldi, line 654: I suspect this isn't going to work as is. `studyVocab` is just going to return a string with the memory address of the `ResponseFuture`. I suspect to make this available as a file in R you'll need to use the RFileSetProcessor in some way or you could just parse the whole thing to a string representation of the R table in Java.

I'll look more tomorrow morning at how the former recommendation would work. I'm not that familiar as of now with our Java -> R communication.

dmgaldi (changes requested):

schema/library.raml
```diff
@@ -394,12 +398,6 @@ types:
       - geoaggregator
       - latitude
       - longitude
-  API_VariableScale:
-    type: string
-    enum:
-      - log
-      - log2
-      - ln
```
dmgaldi, line 402: Looks like a few things got lost in the merge with Ryan's DV code. I don't think this should be removed in addition to some of the other changes in this file.

I think you can just checkout the latest EDA common and re-run `./gradlew gen-jaxrs` to fix this up.

src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
```diff
@@ -611,4 +650,147 @@ public String getRBinListAsString(List<String> labels) {
...
+  public String getRMegastudyAsString(String compressedDataHandle, Map<String, DynamicDataSpecImpl> dataSpecs) {
+    PluginUtil util = getUtil();
+
+    // find and validate variables w study specific vocabs
+    List<DynamicDataSpecImpl> dataSpecsWithStudyDependentVocabs = findDataSpecsWithStudyDependentVocabs(dataSpecs);
+    List<String> entities = dataSpecsWithStudyDependentVocabs.stream().map(data -> getDynamicDataSpecEntityId(data)).toList();
+    boolean allEqualEntities = entities.isEmpty() || Collections.frequency(entities, entities.get(0)) == entities.size();
+    if (!allEqualEntities) { 
+      // TODO get mad 
+    }
```
dmgaldi, line 758: Would it ever be valid to pass duplicate data specs in here? If we don't want to do this validation here we could state that it's a precondition that data specs are not duplicate and let the validation live upstream.

d-callan: hmm. probably we wont ever see duplicate data specs, but im not sure we need to explicitly prevent against it? the validation i was thinking of was more that we actually had to do anything. the circumstances under which we need to impute anything is kind of specific, so it makes sense of those conditions arent met that we simply dont do anything. i just need to remember/ collect a detailed list of what those conditions are..

src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
```diff
@@ -334,6 +336,38 @@ protected Map<String, Double> getCategoricalProportionDistribution(VariableSpec
     return response.getHistogram().stream().collect(Collectors.toMap(HistogramBin::getBinLabel, bin -> Double.valueOf(bin.getValue().toString())));
   }
 
+  protected Map<String, InputStream> getVocabByRootEntity(DynamicDataSpecImpl dataSpec, List<APIFilter> subsetFilters) {
+    PluginUtil util = getUtil();
+    ResponseFuture response = _subsettingClient.getVocabByRootEntity(_referenceMetadata, dataSpec, subsetFilters);
+
+    Map<String, InputStream> vocab = new HashMap<String, InputStream>();
+    try (InputStream vocabStream = response.getInputStream()) {
```
dmgaldi, line 344: I think the problem is that when you assign a closeable variable like an `InputStream` in a "try-with-resources" block like this, it closes the stream at the end of the try block.

src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
```diff
@@ -611,4 +659,155 @@ public String getRBinListAsString(List<String> labels) {
...
+    return megastudyAsRString;
+  }
+
+  public String getRInputDataWithImputedZeroesAsString(String compressedDataHandle, Map<String, DynamicDataSpecImpl> dataSpecs) {
+    boolean validRequest = validateImputeZeroesRequest(dataSpecs);
+    
+    if (validRequest) {
+      String megastudyData = getRMegastudyAsString(compressedDataHandle, dataSpecs);
+      // this feels hacky, but bc this has to be called from the plugin and the plugin already has variables in R... im assuming its present.
+      return "veupathUtils::getDTWithImputedZeroes(" + megastudyData + ", variables)";
```
dmgaldi, line 796: Maybe "variables" could be a parameter so we don't need outside knowledge to understand this method.

src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
```diff
@@ -611,4 +659,155 @@ public String getRBinListAsString(List<String> labels) {
...
+    if (dataSpec.isCollectionSpec()) {
+      return dataSpec.getCollectionSpec().getCollectionId();
+    } else if (dataSpec.isVariableSpec()) {
+      return dataSpec.getVariableSpec().getVariableId();
+    } else {
+      return null;
+    }
+  }
+
+  public String getRMegastudyAsString(String compressedDataHandle, Map<String, DynamicDataSpecImpl> dataSpecs) {
```
dmgaldi, line 770: I'm a little slow at following this. Does this function require certain named inputs to be defined in the R connection that's going to evaluate it?

Can we pass in the R connection and validate that it has all of the prerequisite variables defined?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/CollectionFloatingBoxplotPlugin.java
```diff
@@ -96,19 +96,25 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
...
         spec.getMaxAllowedDataPoints(), 
         "noVariables", 
         nonStrataVarColNames, 
         (name, conn) ->
-        conn.voidEval(util.getVoidEvalFreadCommand(name, inputVarSpecs))
+        conn.voidEval(name + " <- data.table::fread('" + name + "', na.strings=c(''))")
       );
 
+    List<DynamicDataSpecImpl> dataSpecsWithStudyDependentVocabs = findDataSpecsWithStudyDependentVocabs(varMap);
+    Map<String, InputStream> studyVocabs = getVocabByRootEntity(dataSpecsWithStudyDependentVocabs);
```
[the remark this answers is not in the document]
d-callan: There should be a helper in AbstractPlugin.java

src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
d-callan: It needs the tabular data from the merge service and the special vocabs from subsetting loaded, as well as a map of how the variables in the tabular data play in the plot. It puts together a string that represents a r cmd to send to rserve, and that string contains references to those dependencies. If those things aren't present in the r process trying to run the cmd, then rserve will put up an error.

dmgaldi (approved):

PR #317, opened 2023-09-01
R integration attempt

dmgaldi:
Taking a stab at stringifying the study vocabs in R by streaming to rserve.

### DIFF
--- a/src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
+++ b/src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
@@ -41,6 +41,8 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static org.veupathdb.service.eda.common.plugin.util.PluginUtil.singleQuote;
+import static org.veupathdb.service.eda.common.plugin.util.RServeClient.streamResult;
+import static org.veupathdb.service.eda.common.plugin.util.RServeClient.useRConnectionWithRemoteFiles;
 
 /**
  * Base visualization plugin for all other plugins.  Provides access to parts of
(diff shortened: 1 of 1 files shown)


src/main/java/org/veupathdb/service/eda/ds/plugin/AbstractPlugin.java
```diff
@@ -646,22 +648,25 @@ public boolean validateImputeZeroesRequest(Map<String, DynamicDataSpecImpl> data
...
-                                                                                              "variableId=" + getDynamicDataSpecId(dataSpec) + ")" + 
-                                                          "vocabulary=c(',paste(V2, collapse=','),')," +
-                                                          "study=\"',V1,'\"'" +
-                                                          "studyIdColumnName='" + util.getEntityAncestorsAsRVectorString(getDynamicDataSpecEntityId(dataSpec), _referenceMetadata) + "[1])))";
-  
-    String studyVocabsListAsRString = "veupathUtils::StudySpecificVocabulariesByVariable(S4Vectors::SimpleList(paste(" + studyVocabAsRTibble + "[2], collapse=',')))";
-    
-    return studyVocabsListAsRString;
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    useRConnectionWithRemoteFiles(Resources.RSERVE_URL, Map.of("studyVocab", studyVocab.getInputStream()), connection -> {
```
d-callan, line 654: I was intending on calling this helper from the same part of the associated plugin, so that we know we're working in the same r process for the subsequent call to plot.data... I'm wondering if we can pass that connection through to here as a method arg somehow?

dmgaldi: Yeah it's also less overhead to re-use the connection. So would you just want this method to take an R connection as an additional param and assign the vocab data structure to a variable within the R connection?

PR #322, opened 2023-09-25
add empty correlation skeleton

asizemore:
Resolves task in https://github.com/VEuPathDB/web-monorepo/issues/512

Adds the skeleton of a correlation bipartite network viz so that the frontend can actually see a visualization. The goal is to really just let the frontend have the endpoint, but as a byproduct we've gone ahead and set up some of the plugin skeleton.

### DIFF
--- a/schema/library.raml
+++ b/schema/library.raml
@@ -70,6 +70,7 @@ types:
       displayRangeMax?: any
       vocabulary?: string[]
       imputeZero: boolean
+      hasStudyDependentVocabulary?: boolean
       isCollection: boolean
       members?: VariableSpec[]
   ComputedVariableMetadata:
@@ -372,7 +373,7 @@ types:
       isMergeKey: boolean
       isMultiValued: boolean
       imputeZero: boolean
-      hasStudySpecificVocabulary?: boolean
+      hasStudyDependentVocabulary?: boolean
       variableSpecToImputeZeroesFor?: VariableSpec
   API_VariableType:
     type: string
--- a/schema/library.raml
+++ b/schema/library.raml
@@ -70,7 +70,6 @@ types:
       displayRangeMax?: any
       vocabulary?: string[]
       imputeZero: boolean
-      hasStudyDependentVocabulary?: boolean
       isCollection: boolean
       members?: VariableSpec[]
   ComputedVariableMetadata:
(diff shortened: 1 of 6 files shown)


schema/library.raml
```diff
@@ -373,7 +372,7 @@ types:
       isMergeKey: boolean
       isMultiValued: boolean
       imputeZero: boolean
-      hasStudyDependentVocabulary?: boolean
+      hasStudySpecificVocabulary?: boolean
```
d-callan, line 375: whats happened here? this isnt right..

schema/library.raml
```diff
@@ -70,7 +70,6 @@ types:
       displayRangeMax?: any
       vocabulary?: string[]
       imputeZero: boolean
-      hasStudyDependentVocabulary?: boolean
```
d-callan, line 73: boo

schema/library.raml
```diff
@@ -777,6 +776,12 @@ types:
       overlayVariable:
         type: VariableSpec
         required: false
+  CorrelationAssayMetadataBipartitenetworkPostRequest:
+    type: DataPluginRequestBase
+    properties:
+      computeConfig: EmptyDataPluginSpec
+      config: EmptyDataPluginSpec
+  CorrelationAssayMetadataStatsResponse: DifferentialAbundanceStatsResponse
```
d-callan, line 784: do we not know what the response will look like yet?

schema/library.raml
asizemore: i ran `make raml-gen-code` which made this change. Is the original name the correct one? I can remove this change from the PR

schema/library.raml
asizemore: we have a plan but since it doesn't matter for the frontend stub i didn't add it yet

asizemore:
@d-callan do you have any concerns that I need to address before merging? The frontend for correlation is now up but without this piece no one can see. 
I'll be back at the data service in a few days to turn this skeleton into a real plugin

d-callan:
I'd like to know why that new prop is wrong and fix that..

asizemore:
Do you mean `hasStudySpecificVocabulary`? Fixed in the latest commit. I'll resolve the comment above about it so that it's more clear

d-callan (approved):

PR #325, opened 2023-10-16
Add bipartite network for correlation assay vs metadata app

### LINKED ISSUE #299: Add bipartite network plugin
For the correlation app. Expect to receive compute results at the stats endpoint and either pass them straight to the frontend or to plot.data. Still discussing.

asizemore:
Resolves #299 
This PR adds a bipartite network plugin for the correlationAssayMetadata app. The plugin uses the statistics endpoint from the correlation compute plugin, and via plot.data converts these stats into a nice json output that structures the statistics as a bipartite network.


schema/url/correlationassaymetadata/plugin-correlationassaymetadata-bipartitenetwork.raml
```diff
@@ -6,8 +6,22 @@ types:
...
     type: DataPluginRequestBase
     properties:
-      computeConfig: EmptyDataPluginSpec
-      config: EmptyDataPluginSpec
+      computeConfig: CorrelationComputeConfig
+      config: CorrelationAssayMetadataBipartitenetworkSpec
 
   CorrelationAssayMetadataStatsResponse:
-    type: DifferentialAbundanceStatsResponse
+    additionalProperties: false
```
d-callan, line 13: is this specific to assay-metadata? or any correlation?

d-callan: also how does this relate to the CorrelationStatsResponse you picked up from compute? what am i missing?

schema/url/correlationassaymetadata/plugin-correlationassaymetadata-bipartitenetwork.raml
```diff
@@ -6,8 +6,22 @@ types:
...
+    properties:
+      statistics: CorrelationPoint[]
+      data1Metadata: string
+      data2Metadata: string
+
+  CorrelationAssayMetadataBipartitenetworkSpec:
```
d-callan, line 19: similar question.. how is this specific to assay-metadata?

src/main/java/org/veupathdb/service/eda/ds/plugin/correlationassaymetadata/CorrelationAssayMetadataBipartitenetworkPlugin.java
```diff
@@ -45,19 +50,46 @@ protected boolean computeGeneratesVars() {
...
+    useRConnectionWithRemoteFiles(Resources.RSERVE_URL, dataStreams, connection -> {
+      
+      // Take the stats response object and write it into R
+      // as a data.frame. There's probably a much more elegant solution, please advise!
+      connection.voidEval("statsDf <- data.frame(correlationCoef = character(), data1 = character(), data2 = character(), stringsAsFactors=FALSE)");
+      for (int i = 0; i < stats.getStatistics().toArray().length; i++) {
+        CorrelationPoint point = stats.getStatistics().get(i);
+        String newRow = "list(correlationCoef='" + point.getCorrelationCoef() + "', data1='" + point.getData1() + "', data2='" + point.getData2() + "')";
+        connection.voidEval("newRow <- " + newRow);
+        connection.voidEval("statsDf <- rbind(statsDf, newRow)");
```
d-callan, line 79: im a little worried about perf here and wonder if we can find a better way to get the data in R. ill try to think about it some, or we could see if @ryanrdoherty or @dmgaldi have an idea.

src/main/java/org/veupathdb/service/eda/ds/plugin/correlationassaymetadata/CorrelationAssayMetadataBipartitenetworkPlugin.java
```diff
@@ -45,19 +50,46 @@ protected boolean computeGeneratesVars() {
...
+        
+      }
+
+      connection.voidEval("bpNet <- plot.data::bipartiteNetwork(" +
```
d-callan, line 83: i think usually wed have a wrapper sort of method that takes these params, builds the obj and writes to json as a single step. not essential, but might be nice for consistency.

src/main/java/org/veupathdb/service/eda/ds/plugin/correlationassaymetadata/CorrelationAssayMetadataBipartitenetworkPlugin.java
asizemore: it's definitely not fast, but not as bad as i originally feared. Regardless, i think it could use a better approach so yes if any of you have any ideas im all ears!

d-callan: so have you seen this? http://rforge.net/org/doc/  there are a bunch of REXP* classes there, under org.rosuda.REngine that ive known for a while we should probably be looking at using. i wonder if it could help us here and now is a good time to learn more about it?

schema/url/correlationassaymetadata/plugin-correlationassaymetadata-bipartitenetwork.raml
asizemore: updated

asizemore: It's specific to a bipartite network representing correlation data.

schema/url/correlationassaymetadata/plugin-correlationassaymetadata-bipartitenetwork.raml
asizemore: updated

asizemore:
This PR now adds the correlation assay metadata plugin without calling out to R at all. The plugin reshapes the data accordingly. Successfully integration tested.

This is a temporary solution that will allow us to continue making progress on the front end and see the app on qa. Later we can return to the plot.data network classes.

src/main/java/org/veupathdb/service/eda/ds/plugin/correlationassaymetadata/CorrelationAssayMetadataBipartitenetworkPlugin.java
```diff
@@ -45,19 +57,95 @@ protected boolean computeGeneratesVars() {
...
+      link.put("source", sourceNode);
+      link.put("target", targetNode);
+      link.put("strokeWidth", correlationRow.getCorrelationCoef());
+      // Link color is the sign of the correlation
+      String color = Float.parseFloat(correlationRow.getCorrelationCoef()) < 0 ? "-1" : "1";
+      link.put("color", color);
+      links.put(link);
+    });
+
+    // Get unique IDs
```
d-callan, line 126: you might consider using a Set and converting to a List. im not sure the tradeoff in terms of perf, or if we think there will be enough nodes for that to really matter, etc. dan would probably be the person to ask if you wanted to explore it.

d-callan: lol or just using both, something like `if (hashSet.add(E)) arrayList.add(E);`

src/main/java/org/veupathdb/service/eda/ds/plugin/correlationassaymetadata/CorrelationAssayMetadataBipartitenetworkPlugin.java
```diff
@@ -45,19 +57,95 @@ protected boolean computeGeneratesVars() {
...
+    List<String> uniqueColumn1IDs = column1NodeIDs.stream().distinct().collect(Collectors.toList());
+    List<String> uniqueColumn2IDs = column2NodeIDs.stream().distinct().collect(Collectors.toList());
+    List<String> uniqueNodeIDs = nodeIDs.stream().distinct().collect(Collectors.toList());
+
+    // Turn the node ids into an array of Node objects, each with an id property
```
d-callan, line 131: why not just start with a List\<Node> ?

d-callan (commented):
im liking not going to R for this, until we really have to. i dont think the Java+raml is quite there yet though. let me know if the comments are unclear wrt this.

otherwise, i wonder if the java to convert between the stats response from compute to the BipartiteNetworkResponse the client expects can be abstracted away from the plugin? presumably wed want to have the ability to reuse this for the other correlation network vizs..

asizemore:
feedback incorporated! Note that everything that is just the straight data became *data for consistency with other *data classes. I think that's more clear.

schema/url/networks.raml
```diff
@@ -0,0 +1,55 @@
...
+    properties:
+      variables: 
+        type: VariableMapping[]
+        required: false
+
+  BipartiteNetworkConfig:
+    type: NetworkConfig
+    properties:
+      column1Metadata: string
+      column2Metadata: string
```
d-callan, line 35: What're we anticipating this string being? What all metadata do we think is useful here?

d-callan (approved):

PR #326, opened 2023-10-19
Fix quantitative bubble marker plugins

dmgaldi:
## Overview
Lazily load vocabulary to ensure we only try to retrieve it for categorical variables.


ryanrdoherty (approved):

bobular (commented):
Looks good - did you know a general gradle thing snuck in too?

PR #329, opened 2023-11-09
Mbio map viz

d-callan:
this would add a viz to the viz tab, has nothing to do w the standalone map as of now, except to consider this as an alternative to putting a link to the SAM on UD pages


asizemore (approved):

d-callan:
didnt want to merge this before release, just in case, but i see no reason not to now. can be easily undone if needed.

PR #335, opened 2023-11-21
bubbles marker endpoint: only add marker data to output when there is non-null overlayValue

### LINKED ISSUE #334: Bubble Markers - invalid missing overlayValue with certain numerator/denominator combinations
The error is

```
Cannot invoke "java.lang.Double.doubleValue()" because the return value of "org.veupathdb.service.eda.ds.plugin.standalonemap.aggregator.MarkerAggregator.finish()" is null
```

Here is a screenshot:
image

Easiest explained with SCORE S. mansoni with no filters, country variable:

* with Kenya/Kenya, there is an invalid response UNLESS you are zoomed in to a region with NO Ivory Coast or Tanzanian markers
* with Kenya/Kenya+Tanzania, bad response unless you are zoomed in to a region with no Ivory Coast

The problem was that markers were being returned with no `overlayValue` but these markers should have been skipped.

bobular:
Fixes #334 

Builds on https://github.com/VEuPathDB/EdaDataService/commit/d27dea372da9186851c0912e998a6885ffc4d507

### DIFF
--- a/src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/BubbleMapMarkersPlugin.java
+++ b/src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/BubbleMapMarkersPlugin.java
@@ -152,10 +152,12 @@ public class BubbleMapMarkersPlugin extends AbstractEmptyComputePlugin<Standalon
       MarkerAggregator<Double> aggregator = data.getMarkerAggregator();
       if (aggregator != null) {
         Double aggregation = aggregator.finish();
-        if (aggregation != null)
+        if (aggregation != null) {
           mapEle.setOverlayValue(overlayConfig.map(oc -> oc.serializeAverage(aggregation)).orElse(null));
+	  // only output marker data where there are overlay values (issue #334)
+	  output.add(mapEle);
+	}
       }
-      output.add(mapEle);
     }
     StandaloneMapBubblesPostResponse response = new StandaloneMapBubblesPostResponseImpl();
     response.setMapElements(output);


src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/BubbleMapMarkersPlugin.java
```diff
@@ -152,10 +152,12 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
       MarkerAggregator<Double> aggregator = data.getMarkerAggregator();
       if (aggregator != null) {
         Double aggregation = aggregator.finish();
-        if (aggregation != null)
+        if (aggregation != null) {
           mapEle.setOverlayValue(overlayConfig.map(oc -> oc.serializeAverage(aggregation)).orElse(null));
+	  // only output marker data where there are overlay values (issue #334)
+	  output.add(mapEle);
```
ryanrdoherty, line 158: Oh, so if there's no value for this, we don't want to give any of the other info either?

ryanrdoherty: Aha, caught up on this in slack and see why we want it this way.  Approved and merged.

ryanrdoherty (approved):

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/BubbleMapMarkersPlugin.java
bobular: I think so. The (existence of) numerator/denominator values form part of the "little filter" that restricts what markers are shown.

And the front end is expecting a value!

PR #338, opened 2023-11-23
Correlation assay assay

d-callan:
rough draft to get my thoughts down while theyre fresh


api.raml
```diff
@@ -527,15 +527,30 @@ securitySchemes:
...
     displayName: Visualizations for discovering corrleations between assay data and sample metadata
 
     /bipartitenetwork:
       post:
-        description: Returns data required to create a volcanoplot from a differential abundance analysis.
+        description: Returns data required to create a bipartite network from a correlation matrix.
         body:
           application/json:
-            type: lib.CorrelationBipartitenetworkPostRequest
+            type: lib.CorrelationAssayMetadataBipartitenetworkPostRequest
```
d-callan, line 553: this change is the only that has me nervous. do we think this will break anything front end if i merge?

asizemore: i don't think so. If we merge and tag then we can see on my dev site if anything broke. I can then fix any frontend issues tomorrow

PR #339, opened 2023-11-29
filter correlation network on pvalue

d-callan:
add pvalue filter to abstract correlation bipartite network class

### DIFF
--- a/src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
+++ b/src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
@@ -93,7 +93,7 @@ public abstract class AbstractCorrelationBipartiteNetwork<T extends DataPluginRe
       LinkData link = new LinkDataImpl();
       link.setSource(sourceNode);
       link.setTarget(targetNode);
-      link.setStrokeWidth(correlationRow.getCorrelationCoef());
+      link.setStrokeWidth(String.valueOf(Math.abs(Float.parseFloat(correlationRow.getCorrelationCoef()))));
       // Link color is the sign of the correlation
       String color = Float.parseFloat(correlationRow.getCorrelationCoef()) < 0 ? "-1" : "1";
       link.setColor(color);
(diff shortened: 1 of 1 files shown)


d-callan:
tested locally and confirmed working w https://github.com/VEuPathDB/microbiomeComputations/pull/64

d-callan:
again i think ill merge and address feedback later

src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
```diff
@@ -91,7 +95,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
       LinkData link = new LinkDataImpl();
       link.setSource(sourceNode);
       link.setTarget(targetNode);
-      link.setStrokeWidth(correlationRow.getCorrelationCoef());
+      link.setStrokeWidth(String.valueOf(Math.abs(Float.parseFloat(correlationRow.getCorrelationCoef()))));
```
bobular, line 98: I assume the client code scales this to something larger than [0,1]?

Naming it `strokeWidth` suggests that this is used as-is on the client side.  Why not call it `edgeWeight` or something like that? (not urgent, of course)

src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
```diff
@@ -48,6 +48,7 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
 
     CorrelationStatsResponse stats = getComputeResultStats(CorrelationStatsResponse.class);
     Number correlationCoefThreshold = getPluginSpec().getCorrelationCoefThreshold() != null ? getPluginSpec().getCorrelationCoefThreshold() : 0.2;
+    Number pValueThreshold = getPluginSpec().getSignificanceThreshold() != null ? getPluginSpec().getSignificanceThreshold() : 0.05;
```
bobular, line 51: I see that the response from this plugin does not contain any threshold values in `bipartitenetwork.config` - therefore the client doesn't know which default values were used (e.g. 0.2 and 0.05 above) if none were provided in the request.

Is this worth sorting out? I don't know. Just wanted to mention it!

d-callan: i like that suggestion. ill file a ticket for it i think. https://github.com/VEuPathDB/EdaDataService/issues/340

src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
asizemore: good idea bob! Actually by a lucky chance i've added you as a reviewer on the PR that does the scaling. I like your point about how since it gets scaled, it's not really strokeWidth.

PR #342, opened 2023-12-20
Always request all study vocabs

d-callan:
this, together w https://github.com/VEuPathDB/veupathUtils/pull/32, should fix the issue w the histo seen in https://github.com/VEuPathDB/veupathUtils/issues/31

depends on https://github.com/VEuPathDB/EdaCommon/pull/49
depends on https://github.com/VEuPathDB/plot.data/pull/237

needs testing yet

### DIFF
--- a/src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
+++ b/src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
@@ -850,7 +850,7 @@ public abstract class AbstractPlugin<T extends DataPluginRequestBase, S, R> {
         VariableSpecImpl varSpec = new VariableSpecImpl();
         varSpec.setVariableId(var.getVariableId());
         varSpec.setEntityId(entityId);
-        varSpecsWithStudyDependentVocabs.add(var);
+        varSpecsWithStudyDependentVocabs.add(varSpec);
       }
     }
 
@@ -874,7 +874,7 @@ public abstract class AbstractPlugin<T extends DataPluginRequestBase, S, R> {
         CollectionSpecImpl colSpec = new CollectionSpecImpl();
         colSpec.setCollectionId(collection.getCollectionId());
         colSpec.setEntityId(entityId);
-        collectionSpecsWithStudyDependentVocabs.add(collection);
+        collectionSpecsWithStudyDependentVocabs.add(colSpec);
       }
     }
 
--- a/src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
+++ b/src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
@@ -748,8 +748,9 @@ public abstract class AbstractPlugin<T extends DataPluginRequestBase, S, R> {
   // these are deprecated. weve decided to always pass all variables w study dependent vocabs, 
   // rather than only the ones present in the plot. its just going to call the new getDynamicDataSpecsWithStudyDependentVocabs..
   public List<DynamicDataSpec> findDataSpecsWithStudyDependentVocabs(Map<String, DynamicDataSpec> dataSpecs) {
-    // can use the first, we validate theyre all on the same entity
-    String entityId = getDynamicDataSpecEntityId(dataSpecs.get(0));
+    // i dont think this is guaranteed to work properly..... consider actually removing these methods
+    Map.Entry<String, DynamicDataSpec> entry = dataSpecs.entrySet().iterator().next();
+    String entityId = getDynamicDataSpecEntityId(entry.getValue());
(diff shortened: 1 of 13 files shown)


src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
```diff
@@ -836,8 +835,93 @@ public String getRVariableInputDataWithImputedZeroesAsString(String compressedDa
...
+    List<VariableSpec> varSpecsWithStudyDependentVocabs = new ArrayList<>();
+    EntityDef entity = _referenceMetadata.getEntity(entityId).orElseThrow();
+
+    // get reference metadata and find variables w study specific vocabs
+    for (VariableDef var : entity.getVariables()) {
+      if (var.getHasStudyDependentVocabulary()) {
+        VariableSpecImpl varSpec = new VariableSpecImpl();
+        varSpec.setVariableId(var.getVariableId());
+        varSpec.setEntityId(entityId);
+        varSpecsWithStudyDependentVocabs.add(var);
```
bobular, line 853: `add(varSpec)` ?

src/main/java/org/veupathdb/service/eda/ds/core/AbstractPlugin.java
```diff
@@ -836,8 +835,93 @@ public String getRVariableInputDataWithImputedZeroesAsString(String compressedDa
...
+    List<CollectionSpec> collectionSpecsWithStudyDependentVocabs = new ArrayList<>();
+    EntityDef entity = _referenceMetadata.getEntity(entityId).orElseThrow();
+
+    // get reference metadata and find collections w study specific vocabs
+    for (CollectionDef collection : entity.getCollections()) {
+      if (collection.getHasStudyDependentVocabulary()) {
+        CollectionSpecImpl colSpec = new CollectionSpecImpl();
+        colSpec.setCollectionId(collection.getCollectionId());
+        colSpec.setEntityId(entityId);
+        collectionSpecsWithStudyDependentVocabs.add(collection);
```
bobular, line 877: `add(colSpec)`?

bobular (commented):
Looks good apart from the couple of typos I think I spotted.

I didn't understand what the DynamicDataSpecImpl -> DynamicDataSpec changes were about. Just renaming for style reasons?

d-callan:
now tested w boxplots, cleaned up some, ready for final review

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/CollectionFloatingBarplotPlugin.java
```diff
@@ -67,32 +67,37 @@ protected void validateVisualizationSpec(CollectionFloatingBarplotSpec pluginSpe
...
+    List<VariableSpec> plotVariableSpecs = new ArrayList<VariableSpec>();
+    plotVariableSpecs.addAll(pluginSpec.getOverlayConfig().getSelectedMembers());
+
     return ListBuilder.asList(
-      new StreamSpec(DEFAULT_SINGLE_STREAM_NAME, pluginSpec.getOutputEntityId())
-        .addVars(pluginSpec.getOverlayConfig().getSelectedMembers())
-      ); 
+      new StreamSpec(DEFAULT_SINGLE_STREAM_NAME, outputEntityId)
+        .addVars(plotVariableSpecs)
+        // TODO can we make this automagical?
```
ryanrdoherty, line 78: Not clear to me what you want to be automagical?  We probably can with some inheritance if you want to generalize some things.  Just hit me up on slack if you want to discuss.  Looks like there might be a lot of repetition/cut-and-paste in some of these plugins which we would like to avoid.

ryanrdoherty (approved):
Can't say I understand a lot of the interaction between the collection metadata and the R function calls.  It occurs to me that we have an awful lot of R API stuff in here that could maybe be factored into some sort of RPlotClient class, instantiated with the default subset or something?  Would be nice to limit the stuff in AbstractPlugin to just providing an API to get the required data and deliver the generated results back to the web client.

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/CollectionFloatingBarplotPlugin.java
d-callan: there is a lot of repetition, and i hate it, and every time i think about improving the situation it seems to turn into a can of worms.. i start thinking things like 'well in an ideal world id actually be doing some entirely other thing' and things spiral out of control. so if you have ideas, id be open to hearing them

d-callan:
> Can't say I understand a lot of the interaction between the collection metadata and the R function calls

me neither :smile_cat: 

happy to move stuff out of AbstractPlugin. i was also thinking some of that probably is better off in edacommon

Issue #340, opened 2023-12-01
Correlation viz plugins should explicitly return thresholds used


d-callan:
see https://github.com/VEuPathDB/EdaDataService/pull/339#discussion_r1412154669

if no thresholds for correlation coefficient or pvalues are provided, it uses some defaults. the thresholds used arent included in the response, though they should be.

we should address this soonish, definitely before releasing any correlation apps to prod.

PR #343, opened 2024-01-09
return explicit network thresholds

d-callan:
resolves https://github.com/VEuPathDB/EdaDataService/issues/340

are people ok w the revised api?

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -540,7 +540,7 @@ securitySchemes:
           200:
             body:
               application/json:
-                type: lib.BipartiteNetworkPostResponse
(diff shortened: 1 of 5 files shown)


d-callan:
also i think there was another api change we talked about? something about stroke widths vs edge weights. do we have an issue for that? and do we want to handle that at the same time in this pr?

asizemore:
i think it looks good. Makes sense to keep the thresholds separate from the bipartite network.

Yes we were talking about another api change. let me go find that. I'd love to get that in before release. Doesn't have to be this PR if you're ready to merge.

asizemore (approved):
neat!

PR #346, opened 2024-01-16
Remove correlation nans

asizemore:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/755

Removes anything that will be a NaN from our edge list.


d-callan (approved):
I'll just leave a comment that I'm pretty sure this   means that if we decide we want them back for some reason that we'll be changing the API.

PR #350, opened 2024-01-30
Update AppsMetadata.java

dmfalke:
This PR adds VectorBase and PlasmoDB to the pass app and correlation apps (resp).

Note, I added plasmo to both correlation apps. I'm not sure which one is desired, but we can modify that later, if need be.


asizemore (approved):
lgtm!

d-callan:
late to this party.. but i was under the impression wed want to do something similar in host, and figured theyd probably want the non-compute apps in those two places as well? is that a thing worth doing, or talking about now do we think?

PR #353, opened 2024-02-02
Add map markers plugins to apps metadata

### LINKED ISSUE #352: `standalone-map/visualizations/map-markers/bubbles` is missing from `apps` endpoint
I can see the code setting up the (minimal) constraints here:

https://github.com/VEuPathDB/EdaDataService/blob/0d5e8c10c45058218774d964b966055aa293048a/src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/BubbleMapMarkersPlugin.java#L51-L63

But I can't immediately see why it doesn't show up in the `/eda-service/apps` response. Please could you take a look?  I would like to add some constraints and have the front end pick them up and use them (currently the client is using the regular `map-markers` constraints, erroneously!)

dmgaldi:
Resolves #352 

Fix missing marker plugins from apps metadata


src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java
```diff
@@ -56,6 +58,9 @@ public class AppsMetadata {
           "A collection of visualizations designed to support the unbiased exploration of relationships between spatiotemporal variables in a cartographic map.",
           Arrays.asList(VECTORBASE_PROJECT),
           viz("map-markers", new StandaloneMapMarkersPlugin())),
+          viz("map-markers-overlay", new MapMarkersOverlayPlugin()),
```
bobular, line 61: I had to remind myself what all the endpoints do.

`map-markers-overlay` is part of the legacy EDA visualizations for the "pass" app, and doesn't go here.

src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java
```diff
@@ -97,6 +102,8 @@ public class AppsMetadata {
           viz("heatmap", new HeatmapPlugin()),
           viz("map-markers", new MapPlugin()),
           viz("map-markers-overlay", new MapMarkersOverlayPlugin())),
+          viz("map-markers-bubbles", new BubbleMapMarkersPlugin()),
+          viz("map-markers-bubbles-legend", new BubbleMapMarkersLegendPlugin()),
```
bobular, line 106: I don't think these two belong here or for the MBio.Maps section below.

One day maybe (as per Danielle's comment below), but for now I think it will add confusion.

src/main/java/org/veupathdb/service/eda/ds/metadata/AppsMetadata.java
```diff
@@ -146,6 +153,8 @@ public class AppsMetadata {
           // also if we did, id like to consider splitting bubbles into their own viz
           viz("map-markers", new MapPlugin()),
           viz("map-markers-overlay", new MapMarkersOverlayPlugin())),
+          viz("map-markers-bubbles", new BubbleMapMarkersPlugin()),
+          viz("map-markers-bubbles-legend", new BubbleMapMarkersLegendPlugin()),
```
bobular, line 157: See above

bobular (changes requested):
I think the organisation I've suggested is correct!

bobular (approved):
:-)

PR #356, opened 2024-02-14
Optimize the bubble markers to count the negation of filter values set when smaller

dmgaldi:
## Overview
* Optimize the bubble markers to count the negation of the set of filter values if the set contains the majority of values in the vocabulary.
* This is especially useful when using the default configuration of the bubble markers where all values are selected.

### DIFF
--- a/src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/aggregator/CategoricalProportionAggregator.java
+++ b/src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/aggregator/CategoricalProportionAggregator.java
@@ -1,6 +1,8 @@
 package org.veupathdb.service.eda.ds.plugin.standalonemap.aggregator;
 
 import com.google.common.collect.Sets;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.util.Set;
 
@@ -9,6 +11,8 @@ import java.util.Set;
  * set of numerator values over number of elements with values in a set of denominator values.
  */
 public class CategoricalProportionAggregator implements MarkerAggregator<Double> {
+  private static final Logger LOG = LogManager.getLogger(CategoricalProportionAggregator.class);
+
   private final boolean negationMode;
   private final int index;
   private final Set<String> numeratorMatchSet;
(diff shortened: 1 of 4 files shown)


src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/QuantitativeAggregateConfiguration.java
```diff
@@ -1,5 +1,6 @@
 package org.veupathdb.service.eda.ds.plugin.standalonemap.markers;
 
+import com.google.common.collect.Sets;
```
ryanrdoherty, line 3: Extra import

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/aggregator/CategoricalProportionAggregator.java
```diff
@@ -42,9 +53,16 @@ public boolean appliesTo(String[] rec) {
...
   @Override
   public Double finish() {
-    if (numDenominatorMatches == 0) {
+    int totalNumeratorMatches;
+    if (negationMode) {
+      totalNumeratorMatches = numNonNullMatches - numNumeratorMatches;
+    } else {
+      totalNumeratorMatches = numNumeratorMatches;
+    }
+    final int totalDenominatorMatches = totalNumeratorMatches + distinctDenominatorMatchCount;
```
ryanrdoherty, line 62: This seems weird to me, but don't really understand what you're going for.  We could easily be double counting here.  If distinctDenominatorMatchCount is "in denominator set but not in numerator set" (i.e. distinct from numerator matches, then shouldn't you have logic above that does an "else if" after the numeratorMatchSet check?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/QuantitativeAggregateConfiguration.java
```diff
@@ -89,4 +77,49 @@ public String serializeAverage(double d) {
       return Double.toString(d);
     }
   }
+
+  public static class CategoricalProportionAggregatorFactory implements ContinuousAggregators.ContinuousAggregatorFactory {
+    private final CategoricalAggregationConfig categoricalConfig;
+    private final Supplier<List<String>> vocabSupplier;
+    private final Set<String> distinctDenominatorValues;
+    private final Set<String> numeratorMatchSet;
+    boolean negationMode;
```
ryanrdoherty, line 86: Looks like this can also be final.  I assume package for testing?

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/QuantitativeAggregateConfiguration.java
```diff
@@ -89,4 +77,49 @@ public String serializeAverage(double d) {
...
+        this.negationMode = true;
+        this.numeratorMatchSet = numeratorValuesNegation;
+      } else {
+        this.negationMode = false;
+        this.numeratorMatchSet = numeratorValues;
+      }
+      Set<String> denominatorValues = new HashSet<>(categoricalConfig.getDenominatorValues());
+      this.distinctDenominatorValues = Sets.difference(denominatorValues, numeratorValues);
+      this.categoricalConfig = categoricalConfig;
+      this.vocabSupplier = vocabSupplier;
```
ryanrdoherty, line 106: Since you already have the vocab in hand, why not save that off instead of the supplier?  Not clear here if the supplier caches the values; for all we know here it could be expensive to fetch vocab (which is why there's a supplier).

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/QuantitativeAggregateConfiguration.java
```diff
@@ -89,4 +77,49 @@ public String serializeAverage(double d) {
...
+    }
+
+    @Override
+    public MarkerAggregator<Double> create(int index, Function<String, Double> valueQuantifier) {
+      return new CategoricalProportionAggregator(numeratorMatchSet,
+          distinctDenominatorValues,
```
ryanrdoherty, line 112: I see now why you don't have to do the 'else if' above; it's because you already took out the values in the denominator set that live in the numerator set (which is why they're distinct).  But that contract is not clear in the above class.  Maybe add a check above (CategoricalProportionAggregator:30ish) that confirms the intersection is empty.  Or maybe even move the whole "choose negation" logic up there, though don't know if the Aggregator class is used elsewhere- like maybe you need to split it this way for API reasons.  

Seems like the logic for this is spread across these two classes but only by understanding both can you understand what is really going on.

ryanrdoherty (approved):

src/main/java/org/veupathdb/service/eda/ds/plugin/standalonemap/markers/QuantitativeAggregateConfiguration.java
dmgaldi: Yes, at first I had all of the logic in the constructor of the `CategoricalProportionAggregator` class, but it wasn't performant since we instantiate it once per distinct geohash value.

Maybe if the factory and the class live in the same file and the constructor is private it will be more clear.

PR #358, opened 2024-02-15
move correlation type to title and reinstate long subtitle

asizemore:
We recently switched the correlation app titles to be both "Correlation" and the differentiators to be in the subtitle. I didn't realize at the time but turns out on the dashboard view we only show the title (which makes sense). However since we now have two apps wiht the same name...well...
Screen Shot 2024-02-15 at 11 50 42 AM

not so good.

This PR moves the "X v Y" subtitle into the title so that throughout the eda we always know which app is which.


d-callan:
ill approve, but do we want to wait to confirm w dan that it should be functional data vs metadata instead before merging? or even just go ahead and make that change and ask for forgiveness rather than permission should we turn out to be wrong?

d-callan (approved):

asizemore:
I think we should just go ahead, just in case things get down to the wire. Whatever we merge will be better than what we have!

Also i think taxonomic abundance doesn't fall under the functional data umbrella, maybe i'm wrong there, but just in case i'll make that one "Correlation (Taxa, Functional Data v. Metadata)" unless i hear an outcry!

PR #359, opened 2024-02-16
draft node degrees

d-callan:
for bpnet 2.0


src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
```diff
@@ -113,6 +113,8 @@ protected void writeResults(OutputStream out, Map<String, InputStream> dataStrea
     uniqueNodeIDs.forEach((nodeID) -> {
       NodeData node = new NodeDataImpl();
       node.setId(nodeID);
+      int nodeDegree = Collections.frequency(nodeIDs, nodeID);
```
asizemore, line 116: worth noting that this only works because we've ensured the edge list is de-duplicated

asizemore (approved):

asizemore:
right now we're always returning degrees. Ideally we'd ask for degrees and only add them when asked for. I assume this is left as a growth area?

d-callan:
i didnt anticipate that automatically returning them would be an issue tbh. maybe for perf if the network were really quite big, but im always happy to grow into perf concerns. though im also happy to just make it optional if you prefer. so..

PR #360, opened 2024-02-16
Self correlations app

d-callan:
to support a network viz of taxa X taxa correltions for mbio


d-callan:
not tested yet. will get to that next week.

src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/AbstractCorrelationBipartiteNetwork.java
```diff
@@ -33,12 +33,12 @@ public boolean computeGeneratesVars() {
   }
 
   @Override
-  protected void validateVisualizationSpec(CorrelationBipartitenetworkSpec pluginSpec) throws ValidationException {
+  protected void validateVisualizationSpec(CorrelationNetworkSpec pluginSpec) throws ValidationException {
```
asizemore, line 36: I don't follow. Why would we be changing to regular networks here? The object creation at the bottom of this file has partition stuff in it that isn't appropriate for unipartite networks

src/main/java/org/veupathdb/service/eda/ds/plugin/correlation/correlationassayself/CorrelationAssaySelfNetworkPlugin.java
```diff
@@ -0,0 +1,77 @@
...
+  protected void writeResults(OutputStream out, Map<String, InputStream> dataStreams) throws IOException {
+    CorrelationNetworkSpec spec = getPluginSpec();
+    String layout = spec.getLayout() != null ? ", layout = '" + spec.getLayout().getValue() + "'" : "";
+    ByteArrayOutputStream statsBytes = new ByteArrayOutputStream();
+    writeComputeStatsResponseToOutput(out);
+    ByteArrayInputStream statsIn = new ByteArrayInputStream(statsBytes.toByteArray());
+    dataStreams.put("stats-file.tab", statsIn);
+
+    useRConnectionWithRemoteFiles(Resources.RSERVE_URL, dataStreams, connection -> {
+      String command = "veupathUtils::writeJSON(plot.data::Network(data.table::fread(stats-file.tab)" + layout + "))";
```
asizemore, line 73: nit: can we break up this line so that it's easier to read? I'm 100% sure i will mess up the parentheses if i ever edit this file lol

asizemore (commented):
I dont understand the file organization. Why not have one AbstractCorrelationBipartiteNetwork.java and one AbstractCorrelationNetwork.java? The former could even extend the latter.

d-callan:
Yea. maybe i can improve the language some. i think part of the issue is using the word 'network' to sometimes mean a unipartite network specifically, and sometimes as a general or abstract concept.

PR #363, opened 2024-03-19
Query in parallel study vocabularies for different entities

dmgaldi:
## Overview
Looked into slowness in the impute zeroes viz activity. This change avoids invoking the vocab endpoint for each entity in sequence and invokes in parallel instead.


d-callan (approved):
Thanks!

